PR #32, opened 2021-12-22
fix(mgmake): generate makefile with variables for the generated makefiles rather then having to know the path to each file 

viktorvoltaire:
With this change, we support having only namespaced targets. But also makes the entire include process easier.

Before:
```Makefile
include ../.mage/gen.mk

terraform := ../.mage/gen/terraform.mk
$(terraform): $(mage_targets)
	@echo [terraform] generating targets...

include $(terraform)
```

After:
```Makefile
include ../.mage/gen.mk
include $(mg_terraform)
```

This is done by changing so that default file we generate only holds variables to the actual generated makefiles which holds all the targets, it looks like this if multiple namespaces are defined:

```Makefile
mg_targets := /home/volten/git/a-service/.mage/gen/mg-targets.mk
mg_api_gateway := /home/volten/git/a-service/.mage/gen/mg-api-gateway.mk
mg_terraform := /home/volten/git/a-service/.mage/gen/mg-terraform.mk
```


mgmake/generate.go
```diff
@@ -20,24 +21,33 @@ type templateTargets struct {
 
 // GenerateMakefile is a mage target that ...
 func GenerateMakefile(makefile string) error {
+	genDir := filepath.Dir(makefile)
+	mgTargets := filepath.Join(genDir, "mg_targets.mk")
```
odsod, line 25: prefer kebab-case in file names over snake_case

mgmake/generate.go
```diff
@@ -20,24 +21,33 @@ type templateTargets struct {
...
 	mglog.Logger("generate-makefile").Info("generating Makefile...")
 	targets, err := listTargets()
 	if err != nil {
 		return err
 	}
 
+	targettFiles := make(map[string]string)
```
odsod, line 32: spelling: targetFiles

Edholm (approved):
LGTM!

odsod:
I don't understand the commit title and I don't quite understand the examples. What are the file names in the example?

viktorvoltaire:
> I don't understand the commit title and I don't quite understand the examples. What are the file names in the example?

Let me see if I can make it more clear by adding some comments. But basically what ends up happening is we generate a file that looks like below, which is included in `gen.mk` and can then variables can be used by the other includes. Without actually knowing the filename:

```
mg_targets := /home/volten/git/login-service/.mage/gen/mg_targets.mk
mg_apiGateway := /home/volten/git/login-service/.mage/gen/mg_apiGateway.mk
mg_terraform := /home/volten/git/login-service/.mage/gen/mg_terraform.mk
```

mgmake/generate.go
viktorvoltaire: :+1:

odsod:
@viktorvoltaire Allright! Are we missing a transform to snake_case somewhere to make the mg_apiGateway.mk file be mg_api_gateway.mk?

viktorvoltaire:
> @viktorvoltaire Allright! Are we missing a transform to snake_case somewhere to make the mg_apiGateway.mk file be mg_api_gateway.mk?

Yeah, im fixing that as we speak :D

odsod:
IMO we should try and stick to the same casing in the same context, with kebab-case being preferred in e.g. Makefiles and the file system, since it doesn't require holding the shift key down on most keyboard layouts. :smiley_cat:

viktorvoltaire:
@odsod after latest update, we get the following format:

```
mg_targets := /home/volten/git/login-service/.mage/gen/mg-targets.mk
mg_api_gateway := /home/volten/git/login-service/.mage/gen/mg-api-gateway.mk
mg_terraform := /home/volten/git/login-service/.mage/gen/mg-terraform.mk
```

.mage/gen.mk
```diff
@@ -3,6 +3,8 @@ mage_targets_path := $(mage_folder)/gen
 mage_targets := $(mage_targets_path)/targets.mk
 mage := $(mage_targets_path)/local_mage
 
+include $(mage_targets)
```
odsod, line 6: Is there any reason we have `mage_targets` and `mg_targets`? They are hard to tell apart by their name

viktorvoltaire: Yeah I agree, i struggled a bit with what I should call it, so happy to change. `mage_targets` contains the variables that points to the actual makefiles that holds the targets. `mg_targets` is the default one which holds all targets that are not namespaced

odsod: if one contains files and the other contains targets, then how about `mage_files` and `mage_targets`?

odsod: or `mage_target_files`?

odsod: or, if `mage_targets` is the *one magic variable* to always include - maybe we can elevate it and call it `mage_tools` and then to use mage-tools you need to depend on the top-level `mage_tools` variable?

viktorvoltaire: I like `mage_files`. The reason i went for `mg_targets` was mainly because i went with `mg_terraform` etc. But we could go for `mage_terraform`

The variable in `gen.mk` will only be included there.

The root makefile looks something like:
```
.PHONY: all
all: $(mg_targets)

include .mage/gen.mk
include $(mg_targets)
```

So `mage_tools` dsnt really fit it imo. But we could go for your first suggestion?

PR #34, opened 2021-12-22
chore: bump commitlint dependencies

filiptammergard:


Edholm (approved):
Nice! This also requires a newer version of Node iirc? Should be fine though

viktorvoltaire (approved):
:+1:

filiptammergard:
> Nice! This also requires a newer version of Node iirc? Should be fine though

Support for Node 10 is removed, but we should all be running the LTS version (16). Bumping CI as well: https://github.com/einride/mage-tools/pull/35

PR #36, opened 2021-12-23
feat: adds command to generate .mage

viktorvoltaire:
With this, one can run `go run go.einride.tech/mage-tools` to get everything needed to get started with mage-tools

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -1,2 +1,2 @@
 // Package magetools provides utilities and mage targets for building Go projects.
-package magetools
+package main
--- a/example/.mage/magefile.go
+++ b/example/.mage/magefile.go
@@ -1,0 +1,26 @@
+//go:build mage
+// +build mage
+
+package main
+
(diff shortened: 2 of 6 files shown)


Edholm (approved):
damn nice! This will ease the transition immensely!

odsod (approved):
Very nice!

Some thoughts:

* Most tools use `init` as a subcommand for this type of action - worth using the same convention here?
* Do we want an `example` directory and simply `go:embed` the files from there, instead of having inline string constants?

viktorvoltaire:
> * Most tools use `init` as a subcommand for this type of action - worth using the same convention here?
> * Do we want an `example` directory and simply `go:embed` the files from there, instead of having inline string constants?
Yes on both +1

viktorvoltaire:
WDYT @odsod ?

odsod (approved):
Nice!

PR #37, opened 2021-12-23
fix: use git root dir for .tools

viktorvoltaire:
Instead of creating `.tools` in every folder we run mage from, we store everything in the git root directory

### DIFF
--- a/.mage/tools.mk
+++ b/.mage/tools.mk
@@ -1,7 +1,7 @@
 mage_folder := $(abspath $(dir $(lastword $(MAKEFILE_LIST))))
 mage_generated_path := $(mage_folder)/gen
 mage_targets_file := $(mage_generated_path)/targets.mk
-mage := $(mage_generated_path)/local-mage
+mage := $(mage_generated_path)/local-mage 2> /dev/null
 
 include $(mage_targets_file)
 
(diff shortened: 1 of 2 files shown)


.mage/tools.mk
```diff
@@ -1,7 +1,7 @@
 mage_folder := $(abspath $(dir $(lastword $(MAKEFILE_LIST))))
 mage_generated_path := $(mage_folder)/gen
 mage_targets_file := $(mage_generated_path)/targets.mk
-mage := $(mage_generated_path)/local-mage
+mage := $(mage_generated_path)/local-mage 2> /dev/null
```
viktorvoltaire, line 4: I had to add this because of https://github.com/magefile/mage/pull/399 the `sh.Output` triggers `log.Println` and we call this command a lot. We could wait and hope that the PR gets merged, or we can move forward with this solution

viktorvoltaire, line 4: I had to add this because of https://github.com/magefile/mage/pull/399 the `sh.Output` triggers `log.Println` and we call this command a lot. We could wait and hope that the PR gets merged, or we can move forward with this solution

Edholm (approved):
Noice! I read this thinking we started storing `.tools` in the `.git` folder but I see now that it relative to the repo root :+1:

alethenorio (approved):
Good idea

PR #38, opened 2021-12-23
fix(mgtool): use exec.command to solve unwanted logging from global

viktorvoltaire:
After closer debugging, and talking with the mage developers. We get the unwanted logging because the global is set up before logging is configured.


odsod (approved):
LGTM

PR #39, opened 2021-12-23
fix(mgtool): format git root path string to get correct path

viktorvoltaire:
The previous PR did not fail, because technically it did download dependencies to a folder, not the expected one though. Needed to do some string formatting, tested locally and tools now ends up in the correct dir

### DIFF
--- a/mgtool/configuration.go
+++ b/mgtool/configuration.go
@@ -40,7 +40,7 @@ func GetGitRootPath(path string) string {
 	if err := c.Run(); err != nil {
 		panic(err)
 	}
-	return filepath.Join(output.String(), path)
+	return filepath.Join(strings.TrimSpace(output.String()), path)
 }
 
 func GetPath() string {


mgtool/configuration.go
```diff
@@ -40,7 +40,7 @@ func GetGitRootPath(path string) string {
 	if err := c.Run(); err != nil {
 		panic(err)
 	}
-	return filepath.Join(output.String(), path)
+	return filepath.Join(strings.TrimSuffix(output.String(), "\n"), path)
```
odsod, line 43: I typically use strings.TrimSpace for even more fool-proof whitespace trimming

odsod (approved):

mgtool/configuration.go
viktorvoltaire: Ah ofc, that does the job. Also easier to read :) Updated

odsod (approved):
Nice!

PR #40, opened 2021-12-24
fix: generate mgmake import rather then having it in magefile

viktorvoltaire:
The whole `.mage` is built on `mgmake`. We build a binary, then we run the `mgmake` target. So its mandatory, but really only during the generate step. 

Therefor I think it doesnt make sense to have it in the magefiles, since it might not be clear for the end user why it should be there in the first place, removing it would break the entire setup without really explaining why. And it makes the actual magefiles just include the expected targets


Edholm (approved):
LGTM!

.mage/tools.mk
```diff
@@ -2,16 +2,30 @@ mage_folder := $(abspath $(dir $(lastword $(MAKEFILE_LIST))))
 mage_generated_path := $(mage_folder)/gen
 mage_targets_file := $(mage_generated_path)/targets.mk
 mage := $(mage_generated_path)/local-mage
+mgmake := $(mage_folder)/mgmake.go
+
+define mgmake_content
+// +build mage
```
odsod, line 8: do we want a Go 1.17 build directive here too?

example/.mage/tools.mk
```diff
@@ -2,16 +2,30 @@ mage_folder := $(abspath $(dir $(lastword $(MAKEFILE_LIST))))
...
+define mgmake_content
+// +build mage
+
+package main
+
+// mage:import
+import _ "go.einride.tech/mage-tools/mgmake"
+endef
+
+export mgmake_content
```
odsod, line 16: why is this needed?

viktorvoltaire: From what i understood while reading the Makefile documentaiton, it was needed to make `mgmake_content` a usable variable

odsod: ah! is that specific to `define`? (never used that before)

viktorvoltaire: Yeah, first time for me too. I gave it a try without the export, and it seems to be working fine without it :)

odsod: woppa!

.mage/tools.mk
viktorvoltaire: Yep, fixed!

odsod (approved):
Shiny! :star:

viktorvoltaire:
> Shiny! star

One thing, that I now realize while doing a quick test. Is that any dependency that this target brings in, gets added to `go.mod`, meaning if for some reason a user would do a manual `go mod tidy` they would get a diff. It doesnt have an affect when we generate stuff but it could cause some confusion. Rather then deleting it, should we just generate it if it doesnt exist. And add some comments to the file, similar to `wire_gen.go`?

viktorvoltaire:
I pushed my new suggestion :innocent:

example/.mage/tools.mk
```diff
@@ -2,16 +2,30 @@ mage_folder := $(abspath $(dir $(lastword $(MAKEFILE_LIST))))
...
 $(mage_targets_file): $(mage_folder)/go.mod $(mage_folder)/*.go
 	@git clean -fdx $(mage_generated_path)
 	@mkdir -p $(mage_generated_path)
+	$(file > $(mgmake),$(mgmake_content))
 	@cd $(mage_folder) && \
 		go mod tidy && \
 		go run main.go -compile $(mage) && \
-		$(mage) generateMakefile $(@)
+		$(mage) generateMakefile $(@) && \
+		rm -f $(mgmake)
```
odsod, line 28: does this need to be re-generated?

odsod: it shouldn't be here anymore, right?

viktorvoltaire: Right :see_no_evil:

odsod (approved):

PR #41, opened 2021-12-25
chore: Update README.md and update init subcommand

viktorvoltaire:


odsod (commented):
Nice! Can we use the `mgprettier` target to format the own README?

README.md
```diff
@@ -5,5 +5,99 @@ Mage-tools is an opinionated set of [mage](https://github.com/magefile/mage) tar
...
+
+```bash
+go run -mod=mod go.einride.tech/mage-tools -init
+```
+
+Run `make`
+
+## Usage
+
+All mage imports, and targets within the magefiles, gets written to a default Makefile which is always imported. It can also generate specific makefiles to only be imported in certain places like in a Makefile that should only have terraform targets.
```
odsod, line 24: Hmm, are we running with the latest @einride/prettier-config? It should line wrap automatically too.

README.md
```diff
@@ -5,5 +5,99 @@ Mage-tools is an opinionated set of [mage](https://github.com/magefile/mage) tar
...
+
+- [Go](https://golang.org/doc/install) >= 1.17
+- [GNU Make](https://www.gnu.org/software/make/)
+
+## Getting started
+
+To initilize mage-tools in a repository, just run:
+
+```bash
+go run -mod=mod go.einride.tech/mage-tools -init
```
odsod, line 17: IMO it's a bit surprising that `-init` is a flag and not a subcommand. Thinking of e.g. `terraform init`

README.md
```diff
@@ -5,5 +5,99 @@ Mage-tools is an opinionated set of [mage](https://github.com/magefile/mage) tar
...
+package main
+```
+
+#### Imports
+
+If you want to import targets from this repository, just import and add the `// mage:import` comment above the import and alias it with `_` if its only to be used as a mage target
+
+```golang
+// mage:import
+"go.einride.tech/mage-tools/targets/mggo"
```
odsod, line 42: if we encourage a blank line between each `mage:import` I found it becomes more resilient to `goimports` messing with the magic comments - might be superstition though!

README.md
```diff
@@ -5,5 +5,99 @@ Mage-tools is an opinionated set of [mage](https://github.com/magefile/mage) tar
...
+// mage:import
+_ "go.einride.tech/mage-tools/targets/mgsemanticrelease"
+```
+
+#### Local targets
+
+If you wish to utilize an import, or define your own target; like an All target, you can just write one yourself. Just create a public function in a magefile, and it will be included. The target can have no return value other then error.
```
odsod, line 49: let's put `All` in backticks

README.md
viktorvoltaire: I just did that, at least the version we have defined in the target

README.md
viktorvoltaire: Hmm, true

README.md
odsod: Something might be broken in the upstream config! This change should enable prose wrapping: https://github.com/einride/prettier-config/commit/ccf90f0a8d7fb073955daff7e8d54c2c01fd848d

@filiptammergard Not urgent, but something for us to look into.

viktorvoltaire:
> Nice! Can we use the `mgprettier` target to format the own README?

I did, I could actually add it as a target to run in our `All` target

README.md
odsod: For initialization purposes I think it's better to specify a version than `-mod=mod`. I.e.:

`go run go.einride.tech/mage-tools@latest -init`

odsod (approved):
LGTM - some minor comments but nothing blocking!

README.md
viktorvoltaire: Fixed

README.md
viktorvoltaire: Fixed

README.md
viktorvoltaire: Fixed

odsod (approved):
:star_struck:

README.md
filiptammergard: We're only installing @einride/prettier-config in the target, but we're not actually using it. To use it we need to create a `.prettierrc.js` file and register it there. Something like this: https://github.com/einride/mage-tools/pull/43 I didn't get it to work in my PR, probably because I'm not familiar with this way of writing tools, but it's a step in the right direction I think. Perhaps you can see what I missed? @odsod @viktorvoltaire

odsod: @filiptammergard Nice one - I read that PR and I think I know how to get it working - see comment

PR #43, opened 2021-12-26
fix: use @einride/prettier-config when formatting markdown

filiptammergard:
Solves https://github.com/einride/mage-tools/pull/41#discussion_r775265477


targets/mgprettier/targets.go
```diff
@@ -14,17 +14,22 @@ import (
...
+
 var executable string
 
 func FormatMarkdown(ctx context.Context) error {
 	logger := mglog.Logger("prettier")
 	ctx = logr.NewContext(ctx, logger)
-	mg.CtxDeps(ctx, prepare)
+	prettierrc := filepath.Join(mgtool.GetPath(), "prettier", ".prettierrc.js")
+	mg.CtxDeps(ctx, mg.F(prepare, prettierrc))
 	args := []string{
```
odsod, line 33: @filiptammergard I think all that's needed to get it working is to add a flag here that points to the newly written `prettierrc` path

filiptammergard: That's right! Added it and now it formats markdown correctly 👍

odsod: Awesome!!

odsod (approved):
Great stuff!

Edholm (approved):
Nice!

viktorvoltaire:
Awesome @filiptammergard 👌

PR #46, opened 2021-12-27
feat: add mgconvco.ConvcoCheck

odsod:
Faster than commitlint and free of dependency on Node.JS.

### DIFF
--- a/targets/mgconvco/targets.go
+++ b/targets/mgconvco/targets.go
@@ -1,0 +1,59 @@
+package mgconvco
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+
+	"github.com/go-logr/logr"
+	"github.com/magefile/mage/mg"
+	"github.com/magefile/mage/sh"
+	"go.einride.tech/mage-tools/mglog"
+	"go.einride.tech/mage-tools/mgtool"
+)
+
+const version = "0.3.7"
+
+var executable string
+
+func ConvcoCheck(ctx context.Context, rev string) error {
+	logger := mglog.Logger("convco-check")
+	ctx = logr.NewContext(ctx, logger)
+	mg.CtxDeps(ctx, prepare)
+	logger.Info("checking...", "rev", rev)
+	return sh.RunV(executable, "check", rev)
+}
+
+func prepare(ctx context.Context) error {
+	const toolName = "convco"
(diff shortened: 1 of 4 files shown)


Edholm (approved):
Nice! Works roughly the same?

odsod:
@Edholm Yes, maybe a little less strict. But I'm looking into this one too now: https://github.com/cocogitto/cocogitto

odsod:
It's more well-documented than convco, which makes me think it might outlive it. :smile_cat:

targets/mgconvco/targets.go
```diff
@@ -0,0 +1,59 @@
...
+const version = "0.3.7"
+
+var executable string
+
+func ConvcoCheck(ctx context.Context, rev string) error {
+	logger := mglog.Logger("convco-check")
+	ctx = logr.NewContext(ctx, logger)
+	mg.CtxDeps(ctx, prepare)
+	logger.Info("checking...", "rev", rev)
+	return sh.RunV(executable, "check", rev)
```
alethenorio, line 27: This is set on the prepare function but I don't see where it is called. How is the prepare function meant to be called?

alethenorio (approved):
Looks good. Haven't followed the latest changes so uncertain about how this is meant to work.

targets/mgconvco/targets.go
alethenorio: More specifically I meant the "executable" variable

viktorvoltaire: It’s called via the `mg.CtxDeps`

viktorvoltaire (approved):
Awesome!

targets/mgconvco/targets.go
alethenorio: Doh. Missed it somehow

odsod:
@Edholm @alethenorio @viktorvoltaire I added cocogitto as an alterantive git commit checker - it works without any arguments (it reads commits from the latest git tag) and is a little more copy-pasteable (main/master doesn't matter).

Edholm (approved):
Nice!

viktorvoltaire (approved):
Awesome! 👍

alethenorio:
@odsod how does it work for repos that don't have tags?

odsod:
@alethenorio Then it works like without that flag - all commits from HEAD to the first commit

PR #48, opened 2022-01-01
ci: migrate to go-semantic-release

odsod:
Part of removing build-time dependency on Node.js for Go projects.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,6 +26,7 @@ jobs:
         run: make
 
       - name: Release
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


alethenorio (commented):
Doesn't that mean we may end bumping this to v2 on breaking changes?

Edholm (approved):
Noice!

viktorvoltaire (approved):
This is awesome, `allow-initial-development-versions` “Starts your initial development release at 0.1.0” do we need to have this given our release is still pre 1.0.0? Or will it just continue where we’re at

odsod:
@alethenorio Good point! If we are at 0.x I would assume that it still only increments the minor version, but we should verify that - the documentation of go-semantic-release is not super clear.

@viktorvoltaire Good catch - I'm adding that!

PR #50, opened 2022-01-02
fix(init): write to gitignore

viktorvoltaire:
Fixes https://github.com/einride/mage-tools/issues/45

### DIFF
--- a/main.go
+++ b/main.go
@@ -104,6 +104,14 @@ func initMageTools() {
 	if err := execCommandInDirectory(mageDir, "go", []string{"mod", "tidy"}...); err != nil {
 		panic(err)
 	}
+	gitIgnore, err := os.OpenFile(".gitignore", os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0o644)
+	if err != nil {
+		panic(err)
+	}
+	defer gitIgnore.Close()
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -104,6 +104,14 @@ func initMageTools() {
 	if err := execCommandInDirectory(mageDir, "go", []string{"mod", "tidy"}...); err != nil {
 		panic(err)
 	}
+	gitIgnore, err := os.OpenFile(".gitignore", os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0o644)
+	if err != nil {
+		panic(err)
+	}
+	defer gitIgnore.Close()
+	if _, err := gitIgnore.WriteString(".mage/gen/\n.tools/"); err != nil {
```
alethenorio, line 112: Isn't this going to be a problem if the repo already has a .gitignore (it will append to the last line rather than creating a new line)?

Maybe it is enough to add `\n.mage/gen/\n.tools/`?

viktorvoltaire: When testing this locally with and existing file it did append to a new line, so it seems like prefixing with `\n` Isn’t needed if we don’t want an empty line

alethenorio (approved):

PR #53, opened 2022-01-02
chore: make execCmdInDir public

viktorvoltaire:


alethenorio (commented):
@viktorvoltaire what's the reason behind exposing this? Is it for usage in other projects?

viktorvoltaire:
> @viktorvoltaire what's the reason behind exposing this? Is it for usage in other projects?

Yes!

alethenorio:
> > @viktorvoltaire what's the reason behind exposing this? Is it for usage in other projects?
> 
> Yes! 

Don't you think it is simpler to just use exec.Command straight off? exec.Command is a fairly simple function included in the Go stdlib and we are not adding anything extra here but rather just wrapping it.

I think using exec.Command directly would be playing to Go's strength here in readability and simplicity, what do you think?

viktorvoltaire:
> Don't you think it is simpler to just use exec.Command straight off? exec.Command is a fairly simple function included in the Go stdlib and we are not adding anything extra here but rather just wrapping it.
> 
> I think using exec.Command directly would be playing to Go's strength here in readability and simplicity, what do you think?

The whole https://pkg.go.dev/github.com/magefile/mage/sh is a wrapper around exec, which we use everywhere in this repo and in repos consuming it, I figured it be nice for the consumer to expose something that makes it easy to continue with that pattern. 

If not why are we even using the `sh` package 🤔?

viktorvoltaire:
@alethenorio i realised i had missed to push one thing, please have a look again and see if you still have the same opinion. The idea is to make this similar to https://pkg.go.dev/github.com/magefile/mage/sh which we use in this repo and in repos consuming this. Basically to follow the same pattern for executing commands. The `sh` package is also just a wrapper around `exec.Command` so if thats something we should avoid, I think we should rethink using that too?

alethenorio:
I see your point. I honestly don't think the mage package is doing much either though there's enough boilerplate there relates to variable expansion that I buy why it exists.

The changes you made make a little more sense but I wonder how does it differ from just running the mage functions instead, aren't they fulfilling the same goal?

viktorvoltaire:
> I see your point. I honestly don't think the mage package is doing much either though there's enough boilerplate there related to variable expansion that I buy why it exists.
> 
> The changes you made make a little more sense but I wonder how does it differ from just running the mage functions instead, aren't they fulfilling the same goal?

Unfortunately mage does not support running commands in a specific directory. There’s been a PR open for it since May. If they ever do support it, the switch will be easy given this approach imo

alethenorio (approved):

alethenorio:
Got it. Let's go ahead with it then and see if we get any use out of it

PR #54, opened 2022-01-05
feat: enable building Go projects without Node.js dependency

odsod:
Provides alternative to Prettier via markdownfmt


targets/mgmarkdownfmt/targets.go
```diff
@@ -0,0 +1,42 @@
...
+	"path/filepath"
+
+	"github.com/go-logr/logr"
+	"github.com/magefile/mage/mg"
+	"github.com/magefile/mage/sh"
+	"go.einride.tech/mage-tools/mglog"
+	"go.einride.tech/mage-tools/mgtool"
+)
+
+const version = "75134924a9fd3335f76a9709314c5f5cef4d6ddc"
```
viktorvoltaire, line 16: Why are we depending on a commit rather then a version?

odsod: No recent version has been released

targets/mgmarkdownfmt/targets.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+func prepare() error {
+	const toolName = "markdownfmt"
+	toolDir := filepath.Join(mgtool.GetPath(), toolName, version)
+	executable = filepath.Join(toolDir, toolName)
+	if _, err := os.Stat(executable); err == nil {
+		return nil
+	}
+	return sh.RunWithV(
```
viktorvoltaire, line 34: Please use https://github.com/einride/mage-tools/blob/b17068b60126a0460b2c31a1dafa1745a9c2c37d/mgtool/build.go#L13 `GoInstall` instead :) here is an example on when we use it https://github.com/einride/mage-tools/blob/main/targets/mggolangmigrate/targets.go#L25

odsod: Fixed!

viktorvoltaire (approved):
👌

PR #55, opened 2022-01-06
feat(mggolangcilint): use default config with specific linters

odsod:
Based on our current commonly used subset of linters, and preferred Go
version for projects that will integrate this toolchain (>=1.17).


mgtool/build.go
```diff
@@ -10,7 +10,7 @@ import (
 	"github.com/magefile/mage/sh"
 )
 
-func GoInstall(ctx context.Context, goPkg string, version string) (string, error) {
+func GoInstall(ctx context.Context, goPkg, version string) (string, error) {
```
ohlsont, line 13: are we removing this way of writing now?

mgmake/generate.go
```diff
@@ -76,7 +76,7 @@ endif{{end}}
 			return err
 		}
 	}
-	err = os.WriteFile(makefile, createMakefileVariablesFromMap(mgMakefiles), 0o644)
+	err = os.WriteFile(makefile, createMakefileVariablesFromMap(mgMakefiles), 0o600)
```
ohlsont, line 79: I guess in a way this 0o600 could be one global constant maybe with text on why its exactly 0o600.

mgtool/file.go
```diff
@@ -123,6 +123,7 @@ func (s *fileState) handleFileStream(inFile io.Reader, filename string) error {
 		defer out.Close()
 
 		// Write the body to file
+
```
ohlsont, line 126: random new line?

viktorvoltaire (approved):
Really nice to get this default config, great change 👍

mgtool/build.go
odsod: It was auto-fixed by one of the linters

mgmake/generate.go
odsod: It's `gosec` that requires 0600 as permissions for newly created files - IMO octal permissions should be self-explaining (`rw--` in this case)

mgtool/file.go
odsod: Autofixed by linter - probably because the comment above is taken as a "heading", let me see if this can be fixed

mgmake/generate.go
viktorvoltaire: > It's `gosec` that requires 0600 as permissions for newly created files - IMO octal permissions should be self-explaining (`rw--` in this case)
+1

targets/mggolangcilint/golangci.yml
```diff
@@ -0,0 +1,263 @@
...
+    # [fast: false, auto-fix: false]
+    - gosec
+
+    # Find sending http request without context.Context.
+    # [fast: false, auto-fix: false]
+    - noctx
+
+    # Detect missing usage of t.Parallel() method in your Go test.
+    # [fast: true, auto-fix: false]
+    - paralleltest
```
alethenorio, line 203: It is not considered a convention to add t.Parallel in every test I'm Go and some, like hashicorp, even mention avoiding doing it altogether (See https://speakerdeck.com/mitchellh/advanced-testing-with-go?slide=65).

I would rather we don't include this one in the default config

odsod: Good point! I'm submitting a new PR to remove it

odsod: https://github.com/einride/mage-tools/pull/58

PR #57, opened 2022-01-06
feat: move default tools dir under .mage

odsod:
Makes sense to put everything under the .mage dir by default.


viktorvoltaire (approved):
Makes sense 👌

PR #61, opened 2022-01-07
fix: add mage-tools to dependabot config on init

viktorvoltaire:


main.go
```diff
@@ -101,6 +106,69 @@ func initMageTools() {
...
+}
+
+func addToDependabot() error {
+	dependabotYamlPath := filepath.Join(".github", "dependabot.yml")
+	currentConfig, err := ioutil.ReadFile(dependabotYamlPath)
+	if errors.Is(err, os.ErrNotExist) {
+		if err := os.MkdirAll(filepath.Dir(dependabotYamlPath), 0o755); err != nil {
+			return err
+		}
+		err = os.WriteFile(dependabotYamlPath, []byte(dependabotYaml), 0o600)
```
Edholm, line 131: why not embed it as `[]byte` instead of a `string`? :)

Edholm (approved):
Nice!

main.go
```diff
@@ -80,24 +87,27 @@ func initMageTools() error {
...
-		if err := os.WriteFile("Makefile", []byte(makefile), 0o600); err != nil {
+		if err := os.WriteFile("Makefile", makefile, 0o600); err != nil {
 			return err
 		}
 	} else {
-		const mm = "Makefile.MAGE"
-		logger.Info(fmt.Sprintf("Makefile already exist, writing to %s", mm))
-		if err := os.WriteFile(mm, []byte(makefile), 0o600); err != nil {
+		const mm = "Makefile.old"
+		logger.Info(fmt.Sprintf("Makefile already exist, renaming  Makefile to %s", mm))
```
odsod, line 106: spelling: `exists`

targets/mgyamlfmt/targets.go
```diff
@@ -26,14 +25,29 @@ func FormatYaml() error {
 	})
 }
 
+// PreserveEmptyLines adds a temporary #comment on each empty line in the provided byte array.
+// Returns a function that cleans up the temporary comments.
+func PreserveEmptyLines(src []byte) ([]byte, func(in []byte) []byte) {
```
odsod, line 30: why does the cleanup need to be done as a callback closure? why not just a separate function?

odsod: no strong opinion, just looks a little out of the ordinary

odsod (approved):

targets/mgyamlfmt/targets.go
viktorvoltaire: Fixed

targets/mgyamlfmt/targets.go
```diff
@@ -26,14 +25,31 @@ func FormatYaml() error {
 	})
 }
 
+// PreserveEmptyLines adds a temporary #comment on each empty line in the provided byte array.
+// CleanupPreserveEmptyLines can be used to clean up the temporary comments.
+func PreserveEmptyLines(src []byte) []byte {
```
odsod, line 30: can be private?

viktorvoltaire: Its public so I can also use it when appending to dependabot from the init command

odsod: Ahh!

PR #62, opened 2022-01-07
feat: add yaml format target

viktorvoltaire:
Since we are adding yaml as a dependency in https://github.com/einride/mage-tools/pull/61

### DIFF
--- a/targets/mgyamlfmt/targets.go
+++ b/targets/mgyamlfmt/targets.go
@@ -1,0 +1,54 @@
+package mgyamlfmt
+
+import (
+	"bytes"
+	"io/fs"
+	"io/ioutil"
+	"path/filepath"
+	"regexp"
+	"strings"
+
+	"go.einride.tech/mage-tools/mglog"
+	"go.einride.tech/mage-tools/mgpath"
+	"gopkg.in/yaml.v3"
+)
+
+func FormatYaml() error {
+	logger := mglog.Logger("format-yaml")
+	logger.Info("formatting yaml files...")
+	return filepath.WalkDir(mgpath.FromGitRoot("."), func(path string, d fs.DirEntry, err error) error {
+		if filepath.Ext(path) == ".yml" || filepath.Ext(path) == ".yaml" {
+			if err := formatFile(path); err != nil {
+				return err
+			}
+		}
+		return nil
+	})
+}
+
+func formatFile(path string) error {
+	node := yaml.Node{}
+	yamlFile, err := ioutil.ReadFile(path)
+	if err != nil {
+		return err
+	}
+	// Insert temporary comment to preserve empty lines.
+	yamlFile = bytes.ReplaceAll(yamlFile, []byte("\n\n"), []byte("\n#preserveEmptyLine\n"))
+	if err := yaml.Unmarshal(yamlFile, &node); err != nil {
+		return err
+	}
+	var b bytes.Buffer
+	encoder := yaml.NewEncoder(&b)
+	encoder.SetIndent(2)
+	if err := encoder.Encode(&node); err != nil {
+		return err
+	}
+	// Remove temporary comment.
+	indentPreserveComment := regexp.MustCompile("\n\\s+#preserveEmptyLine\n")
+	out := b.String()
+	out = indentPreserveComment.ReplaceAllString(out, "\n\n")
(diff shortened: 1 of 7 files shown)


targets/mgyamlfmt/targets.go
```diff
@@ -0,0 +1,40 @@
...
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"go.einride.tech/mage-tools/mgtool"
+	"gopkg.in/yaml.v3"
+)
+
+func FormatYaml() error {
+	return filepath.Walk(mgtool.GetGitRootPath("."), func(path string, _ os.FileInfo, err error) error {
```
viktorvoltaire, line 13: Should we support excluding directories ? Right now im just defaulting to the git root

.github/dependabot.yml
```diff
@@ -1,13 +1,12 @@
 version: 2
-
```
viktorvoltaire, line 2: Not a 100% sure about this format tbh, feels like blank lines should be kept?

viktorvoltaire: Haven’t looked into https://pkg.go.dev/gopkg.in/yaml.v3#Style might be something there

viktorvoltaire: Seems like there isnt an option for this

viktorvoltaire:
I pushed a new commit to preserve blank lines. It does the job, in a somewhat hacky manner.

.github/dependabot.yml
viktorvoltaire: See my latest commit for a potential solution

targets/mgyamlfmt/targets.go
```diff
@@ -0,0 +1,55 @@
...
+
+	"go.einride.tech/mage-tools/mglog"
+	"go.einride.tech/mage-tools/mgtool"
+	"gopkg.in/yaml.v3"
+)
+
+func FormatYaml() error {
+	logger := mglog.Logger("format-yaml")
+	logger.Info("formatting yaml files...")
+	return filepath.Walk(mgtool.GetGitRootPath("."), func(path string, _ fs.FileInfo, err error) error {
```
Edholm, line 19: btw I think [`filepath.WalkDir`](https://pkg.go.dev/path/filepath#WalkDir) is the recommended func nowadays due to it now having to call `stat` on _every_ file. No need to change, just a heads-up

Edholm (approved):
Noice work!

.github/dependabot.yml
odsod: :star_struck:

odsod:
Files ending in multiple blank lines might be able to be handled by doing `bytes.TrimSpace` and then appending a single trailing endline.

viktorvoltaire:
> Files ending in multiple blank lines might be able to be handled by doing `bytes.TrimSpace` and then appending a single trailing endline.

Ah true!

targets/mgyamlfmt/targets.go
viktorvoltaire: Fixed

targets/mgyamlfmt/targets.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+	var b bytes.Buffer
+	encoder := yaml.NewEncoder(&b)
+	encoder.SetIndent(2)
+	if err := encoder.Encode(&node); err != nil {
+		return err
+	}
+	// Remove temporary comment.
+	indentPreserveComment := regexp.MustCompile("\n\\s+#preserveEmptyLine\n")
+	out := b.String()
```
odsod, line 48: @viktorvoltaire Minor thought: The conversion to string and back to bytes here should be avoidable. (Converting back and forth allocates memory for a copy).

viktorvoltaire: Ah, will have a look and see if I can fix that 👍

viktorvoltaire: @odsod actually I already fixed that. Was annoyed with it 😊 https://github.com/einride/mage-tools/commit/ee6ea005bfd2721fd1165f58320a72b09ec3fe0c#diff-bd1f7610004a598ff323f80c6917c7ede5455328026fc0397bc78b7554a93d74R38

odsod: Sweet

PR #64, opened 2022-01-10
feat(hadolint): add Dockerfile linter

Edholm:
See https://github.com/hadolint/hadolint for more information.

No default config has been specified since I don't think we are using
a custom default config right now. Can be added in the future if needed.

### DIFF
--- a/targets/mghadolint/targets.go
+++ b/targets/mghadolint/targets.go
@@ -1,0 +1,68 @@
+// Package mghadolint is a Dockerfile linter that you can employ to check for Dockerfile best
+// practices and common mistakes.
+package mghadolint
+
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+	"runtime"
+	"strings"
+
+	"github.com/go-logr/logr"
+	"github.com/magefile/mage/mg"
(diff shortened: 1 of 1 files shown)


targets/mghadolint/targets.go
```diff
@@ -0,0 +1,68 @@
...
+const version = "2.8.0"
+
+// nolint: gochecknoglobals
+var executable string
+
+func Hadolint(ctx context.Context) error {
+	ctx = logr.NewContext(ctx, mglog.Logger("hadolint"))
+	mg.CtxDeps(ctx, mg.F(prepare))
+	logr.FromContextOrDiscard(ctx).Info("running...")
+	dockerfilesRaw, err := sh.Output("git", "ls-files", "--exclude-standard", "--cached", "--others", "--", "*Dockerfile")
```
odsod, line 29: should it be `*Dockerfile*`? (Trailing wildcard for e.g. `Dockerfile.armv5`?

Edholm: Fixed @odsod

odsod (approved):

PR #65, opened 2022-01-10
feat(golangcilint): disable `forcetypeassert` linter

Edholm:
this linter has a bit too many false positives where the assetion has
been checked beforehand. Let's remove it for now and re-add it for a
stricter linter in the future.


odsod (approved):
change to `feat` to trigger a release?

PR #66, opened 2022-01-10
feat: make mage-tools easy to understand

viktorvoltaire:
Todo:
- [x]  support specific makefiles not just the root one, like the ones in terraform or in proto folders


.mage/magefile.go
```diff
@@ -25,6 +25,14 @@ import (
 	"go.einride.tech/mage-tools/targets/mggoreview"
 )
 
+func init() {
+	mgmake.MakeFiles[ThisGoesToSeparateMakefile{}] = mgmake.Makefile{
```
viktorvoltaire, line 29: This is my suggestion for how we can specify unique files for certain namespaces, thoughts?
you can see that it got generated https://github.com/einride/mage-tools/pull/66/files#diff-c79aed384364df14b332ed00088de2dfda4d81b327bfec65e680dc01937e0347

odsod: Two thoughts spontaneously:

* Would rather have a method call than explicitly assigning to a global map
* Still not clear on how the working directory works when the Makefile is written into .mage - will `.mage` be the working directory?

.mage/magefile.go
```diff
@@ -25,15 +27,36 @@ import (
...
+			{
+				Path:          mgtool.GetGitRootPath("proto/Makefile"),
+				Namespace:     Proto{},
+				DefaultTarget: "Hej",
+			},
+		},
+	)
+}
+
+var Default = All
```
odsod, line 42: Maybe we should just call it `Default` to avoid having this extra variable?

.mage/magefile.go
```diff
@@ -25,15 +27,36 @@ import (
 	"go.einride.tech/mage-tools/targets/mggoreview"
 )
 
+func init() {
+	mgmake.AddExtraMakefiles(
+		[]mgmake.Makefile{
+			{
+				Path:          mgtool.GetGitRootPath("proto/Makefile"),
```
odsod, line 34: spontaneous naming thought - `mgpath.FromGitRoot`?

.mage/magefile.go
```diff
@@ -25,15 +27,36 @@ import (
 	"go.einride.tech/mage-tools/targets/mggoreview"
 )
 
+func init() {
+	mgmake.AddExtraMakefiles(
+		[]mgmake.Makefile{
```
odsod, line 32: make this method a var-args so we don't have to create an inline slice?

.mage/magefile.go
odsod: Or maybe setting the `.DEFAULT_TARGET` is done by a method call to the `mgmake` package instead of a magic variable?

.mage/magefile.go
viktorvoltaire: Fixed a method call, path is now explained more in the example

.mage/magefile.go
viktorvoltaire: We could change the entire logic, and instead of automatically creating a `Makefile` in the root of the git repo, that file too needs to be defined, similar to how we define the proto makefile? That would potenially make stuff a bit more clear too the end user?

.mage/magefile.go
viktorvoltaire: Ah, I like it!

.mage/magefile.go
```diff
@@ -25,15 +27,36 @@ import (
...
+			{
+				Path:          mgtool.GetGitRootPath("proto/Makefile"),
+				Namespace:     Proto{},
+				DefaultTarget: "Hej",
```
Edholm, line 36: is `Hej` correct here or just a test? :D

mgmake/generate.go
```diff
@@ -18,77 +21,132 @@ type templateTargets struct {
...
+		}
+		namespace := reflect.TypeOf(i.Namespace).Name()
+		makefiles[toMakeTarget(namespace)] = i
 	}
-	mglog.Logger("generate-makefile").Info("generating Makefile...")
+}
+
+const defaultNamespace = "default"
+
+// GenerateMakefiles is a mage target that ...
```
Edholm, line 51: > that ...

should we fill it out before merge mayhaps?

Edholm (commented):
Looks good to me! I agree with the other comments

.mage/magefile.go
viktorvoltaire: Yeah, this was mainly to showcase the PR. Will remove before we merge

mgmake/generate.go
viktorvoltaire: Sure!

.mage/magefile.go
```diff
@@ -25,15 +27,34 @@ import (
 	"go.einride.tech/mage-tools/targets/mggoreview"
 )
 
+func init() {
+	mgmake.AddExtraMakefiles(
```
odsod, line 31: As discussed, let's call it `GenerateMakefiles` to make it more self-describing what the method does

Makefile
```diff
@@ -1,5 +1,39 @@
+# Code generated by Mage-tools. DO NOT EDIT.
```
odsod, line 1: minor: let's have blank lines in the generated file that will make it indistinguishable from a hand-written file - i.e. blank line after this top-level comment, and a blank line after `.DEFAULT_GOAL`

main.go
```diff
@@ -47,11 +48,11 @@ func initMageTools() {
 		}
 	}()
 
-	if mgtool.GetCWDPath("") != mgtool.GetGitRootPath("") {
+	if mgpath.FromCWD("") != mgpath.FromGitRoot("") {
```
odsod, line 51: IMO would be nice to avoid a 3-letter abbreviation in the API here - how about `mgpath.FromWorkDir`?

main.go
```diff
@@ -47,11 +48,11 @@ func initMageTools() {
...
 		panic(fmt.Errorf("can only be generated in git root directory"))
```
odsod, line 52: Use `log.Panicf` or `log.Fatalf` instead of `panic(fmt.Errorf())`

proto/Makefile
```diff
@@ -0,0 +1,8 @@
+# Code generated by Mage-tools. DO NOT EDIT.
+.DEFAULT_GOAL := hej
+include ../.mage/tools.mk
+
+.PHONY: hej
+hej: $(mage)
+	@$(mage) proto:hej
+
```
odsod, line 8: Would be nice to avoid outputting this trailing blank line

main.go
viktorvoltaire: Im using panic because i have a defer func that hadles it, is that bad practice?
```go
	defer func() {
		if err := recover(); err != nil {
			logger.Error(err.(error), err.(error).Error())
		}
	}()
```

mgmake/generate.go
```diff
@@ -18,77 +21,142 @@ type templateTargets struct {
...
+			panic("Path needs to be defined")
+		}
+		namespace := defaultNamespace
+		if i.Namespace != nil {
+			namespace = reflect.TypeOf(i.Namespace).Name()
+		}
+		defaultTarget := runtime.FuncForPC(reflect.ValueOf(i.DefaultTarget).Pointer()).Name()
+		defaultTarget = strings.TrimPrefix(defaultTarget, "main.")
+		defaultTarget = strings.TrimPrefix(defaultTarget, namespace+".")
+		defaultTarget = strings.Split(defaultTarget, "-")[0]
```
viktorvoltaire, line 52: @odsod thoughts on this? The reason behind it is because the function names are originally `main.All` & `main.Proto.Hej-fm` 
Would like a better solution for extracting the function name but this is what i came up with

viktorvoltaire:
@odsod off topic, the checkout stratergy seems a bit weird for this project. It got an error based on something that exist on the master branch, but its not on my local branch. Do we merge to master on checkout? Feels strange :thinking:

Makefile
viktorvoltaire: Fixed

main.go
viktorvoltaire: Yeah thats better, fixed :+1:

proto/Makefile
viktorvoltaire: Fixed

mgmake/generate.go
```diff
@@ -1,94 +1,168 @@
...
+		makefiles[toMakeTarget(namespace)] = makefile{Path: i.Path, DefaultTarget: defaultTarget}
+	}
+}
+
+const defaultNamespace = "default"
+
+func GenMakefiles() error {
+	if len(makefiles) == 0 {
+		return fmt.Errorf("no makefiles to generate, " +
+			"provide some info about adding an init method or linking to readme")
```
viktorvoltaire, line 63: We need a better message here :P

.mage/magefile.go
```diff
@@ -25,15 +27,36 @@ import (
...
+func init() {
+	mgmake.GenerateMakefiles(
+		mgmake.Makefile{
+			Path:          mgpath.FromGitRoot("Makefile"),
+			DefaultTarget: All,
+		},
+		mgmake.Makefile{
+			Path:          mgpath.FromGitRoot("proto/Makefile"),
+			Namespace:     Proto{},
+			DefaultTarget: Proto{}.Hej,
```
odsod, line 39: What happens if the user provides `Proto.Hej` instead of `Proto{}.Hej`?

example/.mage/rules.mk
```diff
@@ -0,0 +1,10 @@
+mage_folder := $(abspath $(dir $(lastword $(MAKEFILE_LIST))))
+mage_tools_path := $(mage_folder)/tools
+mage := $(mage_tools_path)/mgmake/magefile
+
+$(mage): $(mage_folder)/go.mod $(mage_folder)/*.go
+	@go run main.go gen
+
+.PHONY: mage-clean
+mage-clean:
+	@git clean -fdx $(mage_folder)
```
odsod, line 10: nit: missing trailing endline here (as GitHub points out)

Makefile
```diff
@@ -1,5 +1,41 @@
...
+
+.DEFAULT_GOAL := all
+
+include .mage/rules.mk
```
odsod, line 5: Let's keep the name as `.mage/tools.mk`?

mgtool/mgpath/path.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"bytes"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+)
+
+// Path This should only be used to set a custom value.
```
odsod, line 11: This comment stutters `Path This`

mgtool/mgpath/path.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+func Tools() string {
+	if mgToolsPath == "" {
+		panic("No tools path set")
+	}
+	return mgToolsPath
+}
+
+func SetTools(p string) {
```
odsod, line 48: IMO I think we should not make this configurable until there's a compelling reason to do it!

mgmake/generate.go
odsod: As long as it supports also `Proto.Hej` and it throws an error if the parsing turns up something unexpected, then let's go for it!

.mage/magefile.go
viktorvoltaire: `Cannot use 'Proto.Hej' (type func(Proto)) as the type func()`

odsod: Change the type to `interface{}`? This won't work if the target returns `error` or takes a `context.Context` anyway?

viktorvoltaire: Sounds good :+1:

mgmake/generate.go
viktorvoltaire: Sounds good, added verification to make sure that the end result only contains letters

main.go
```diff
@@ -32,75 +46,87 @@ func main() {
 
 	switch os.Args[1] {
 	case "init":
-		initMageTools()
+		if err := initMageTools(); err != nil {
+			panic(err)
+		}
+	case "gen":
```
viktorvoltaire, line 52: @odsod thoughts right now i have not exposed `gen` in the help, meaning its "invisible" for the user, thoughts?

example/.mage/tools.mk
```diff
@@ -1,49 +1,11 @@
 mage_folder := $(abspath $(dir $(lastword $(MAKEFILE_LIST))))
```
odsod, line 1: minor: let's prefer `dir` over `folder` for consistency?

example/.mage/tools.mk
```diff
@@ -1,49 +1,11 @@
...
-mage_generated_path := $(mage_folder)/tools/mgmake
 mage_tools_path := $(mage_folder)/tools
-mage_targets_file := $(mage_generated_path)/targets.mk
-mage := $(mage_generated_path)/local-mage
-mgmake := $(mage_folder)/mgmake_gen.go
-mgmain := $(mage_folder)/mgmain_gen.go
+mage := $(mage_tools_path)/mgmake/magefile
+version := $(shell cd $(mage_folder) && go list -m -f '{{ .Version }}' go.einride.tech/mage-tools)
```
odsod, line 4: let's call this `mage_version` to avoid naming collissions when this file is `include`:d

example/.mage/tools.mk
```diff
@@ -1,49 +1,11 @@
...
-	@git clean -fdx $(mage_generated_path)
-	@mkdir -p $(mage_generated_path)
-	$(file > $(mgmake),$(mgmake_content))
-	$(file > $(mgmain),$(mgmain_content))
-	@cd $(mage_folder) && \
-		go mod tidy && \
-		go run $(notdir $(mgmain)) -compile $(mage) && \
-		$(mage) generateMakefile $(@)
+$(mage): $(mage_folder)/go.mod $(mage_folder)/*.go
+	@go run go.einride.tech/mage-tools@$(version) gen
```
odsod, line 7: the `@version` should not be needed if this is run from the `.mage` directory, the version should be picked up by `go.mod`

viktorvoltaire: Fixed

example/.mage/tools.mk
viktorvoltaire: As we mentioned below, version is not needed

example/.mage/tools.mk
viktorvoltaire: Fixed

main.go
viktorvoltaire: Goit it, fixed

Edholm (commented):
Looks good I think! Just some minor cleanup and we're ready for merge?

.mage/magefile.go
```diff
@@ -25,15 +27,36 @@ import (
...
+	)
+}
+
 func All() {
 	mg.Deps(
 		mgcocogitto.CogCheck,
 		mggolangcilint.GolangciLint,
 		mggoreview.Goreview,
 		mgmarkdownfmt.FormatMarkdown,
+		Proto.Hej,
```
odsod, line 50: we should clean this up before merging

.mage/tools.mk
```diff
@@ -1,49 +1,10 @@
-mage_folder := $(abspath $(dir $(lastword $(MAKEFILE_LIST))))
-mage_generated_path := $(mage_folder)/tools/mgmake
-mage_tools_path := $(mage_folder)/tools
-mage_targets_file := $(mage_generated_path)/targets.mk
-mage := $(mage_generated_path)/local-mage
-mgmake := $(mage_folder)/mgmake_gen.go
-mgmain := $(mage_folder)/mgmain_gen.go
+mage_dir := $(abspath $(dir $(lastword $(MAKEFILE_LIST))))
+mage_tools_path := $(mage_dir)/tools
+mage := $(mage_tools_path)/mgmake/magefile
```
odsod, line 3: is the runnable binary supposed to be named `magefile` and not `mage`?

odsod: now I get it! because it is not `mage` - it is our compiled magefile(s)!

main.go
odsod: no strong opinion here :+1:

main.go
```diff
@@ -17,90 +20,102 @@ var (
...
 	logger.Info("initializing mage-tools...")
-	defer func() {
-		if err := recover(); err != nil {
-			logger.Error(err.(error), err.(error).Error())
-		}
-	}()
 
-	if mgtool.GetCWDPath("") != mgtool.GetGitRootPath("") {
-		panic(fmt.Errorf("can only be generated in git root directory"))
+	if mgpath.FromWorkDir("") != mgpath.FromGitRoot("") {
```
odsod, line 75: would there be any difference between "." and "" here?

odsod: "." would be more clear IMO

mgmake/generate.go
odsod: not specifying any makefiles to generate won't trigger this error, right?

odsod (approved):
N I C E :framed_picture: (please still resolve remaning comments)

main.go
viktorvoltaire: No difference, both works

mgmake/generate.go
viktorvoltaire: If you try and run `GenMakefiles` without having an `init` method where you added makefiles in the magefile, this error will trigger

PR #67, opened 2022-01-13
fix(example): use convco instead of cocogitto

viktorvoltaire:


Edholm (approved):
Noice! Should we remove convco or is there a reason to keep it?

viktorvoltaire:
> Noice! Should we remove convco or is there a reason to keep it?

Lets keep it, we ran into some issues with it but might migrate back in the future

PR #74, opened 2022-01-16
feat: remove tools.mk dependency

viktorvoltaire:
Since we generate makefiles. there's no real reason to keep `tools.mk` around


odsod (approved):
Bloody brilliant! :sparkler:

Makefile
```diff
@@ -2,7 +2,15 @@
...
+
+$(magefile): $(mage_dir)/go.mod $(shell find $(mage_dir)/.. -type f -name '*.go')
+	@cd $(mage_dir) && go run go.einride.tech/mage-tools gen
+
+.PHONY: mage-clean
+mage-clean:
+	@git clean -fdx $(mage_dir)
 
 .PHONY: all
 all: $(mage)
```
odsod, line 16: the `$(mage)` here should also get updated?

viktorvoltaire: Fixed

odsod (approved):
Let's go :raised_hands:

PR #76, opened 2022-01-17
fix: move build into mgmake package

viktorvoltaire:


.mage/magefile.go
```diff
@@ -31,8 +31,8 @@ import (
 )
 
 func init() {
-	mgmake.GenerateMakefiles(
-		mgmake.Makefile{
+	gen.GenerateMakefiles(
```
odsod, line 34: IMO we should find a better package name for this public API

viktorvoltaire: I reverted to the old package name

example/.mage/magefile.go
```diff
@@ -5,7 +5,7 @@ package main
 
 import (
 	"github.com/magefile/mage/mg"
-	"go.einride.tech/mage-tools/mgmake"
+	"go.einride.tech/mage-tools/mgbuild/mgmake"
```
odsod, line 8: does `mgmake` need to be a child, or can we maintain the flat structure by putting `mgmake` at the same level?

mgbuild/main.go
```diff
@@ -0,0 +1,41 @@
+package main
```
odsod, line 1: Non top-level commands are usually put under the `cmd` subdirectory

odsod: `go run go.einride.tech/mage-tools/cmd/build`

example/.mage/magefile.go
viktorvoltaire: It does not

mgbuild/main.go
viktorvoltaire: :+1:

example/.mage/magefile.go
odsod: Disregard me - since it's a command let's go with the comment below!

mgbuild/main.go
viktorvoltaire: Fixed

PR #77, opened 2022-01-17
fix(init): add missing import during init

viktorvoltaire:


odsod (commented):
Why do we need two separate generated Go files? Can we combine this into the same generated file?

viktorvoltaire:
> Why do we need two separate generated Go files? Can we combine this into the same generated file?

2 reasons, 1 being `go:embed` doesn't seem to allow for `../..` in path and the other one because the package in the `init` one is `gopkg.in/yaml.v3` which potentially is only a dependency during init. So we get a diff when running `go mod tidy` since we are removing the temporary files. If we leave them checked in its not a problem.

 Sure if the a user adds `mgyamlfmt` they will pull this dependency in, but we should force users to have that

odsod (approved):

PR #78, opened 2022-01-17
feat: add prepare target, option to depend on prepare and start symlinking to bin dir

viktorvoltaire:
- chore: add prepare function to each target
- feat: add option to depend on prepare

### DIFF
--- a/mgtool/build.go
+++ b/mgtool/build.go
@@ -12,15 +12,15 @@ import (
 )
 
 func GoInstall(ctx context.Context, goPkg, version string) (string, error) {
-	toolDir, err := filepath.Abs(mgpath.Tools())
-	if err != nil {
-		return "", err
-	}
-	executable := filepath.Join(toolDir, goPkg, version, "bin", filepath.Base(goPkg))
+	executable := filepath.Join(mgpath.Tools(), "go", goPkg, version, "bin", filepath.Base(goPkg))
+	symlink := filepath.Join(mgpath.Tools(), mgpath.BinDir, filepath.Base(goPkg))
 
 	// Check if executable already exist
 	if _, err := os.Stat(executable); err == nil {
-		return executable, nil
+		if err := createSymlink(executable, symlink); err != nil {
+			return "", err
+		}
+		return symlink, nil
 	}
 	goPkgVer := fmt.Sprintf("%s@%s", goPkg, version)
 	os.Setenv("GOBIN", filepath.Dir(executable))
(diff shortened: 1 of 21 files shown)


mgtool/build.go
```diff
@@ -12,21 +12,36 @@ import (
...
 	}
 	goPkgVer := fmt.Sprintf("%s@%s", goPkg, version)
 	os.Setenv("GOBIN", filepath.Dir(executable))
 	logr.FromContextOrDiscard(ctx).Info(fmt.Sprintf("Building %s...", goPkgVer))
 	if err := sh.RunV("go", "install", goPkgVer); err != nil {
 		return "", err
 	}
-	return executable, nil
+	if err := createSymlink(executable, symlink); err != nil {
+		return "", err
```
Edholm, line 32: maybe wrap this err to make it easier to troubleshoot:

```suggestion
		return "", fmt.Errorf("failed to create symlink: %w", err)
```

Edholm (approved):
Looks good! :+1: 

What is the reason for the change though? :thinking:

odsod:
@Edholm We want to be able to prepare tools without using them from mage - for example when they are used by Buf, or `go generate`.

mgtool/build.go
odsod: IMO the `createSymlink` function should talk about failure to add symlink

targets/mgconvco/targets.go
```diff
@@ -21,6 +21,12 @@ const version = "0.3.7"
 // nolint: gochecknoglobals
 var executable string
 
+type Prepare mgtool.Prepare
+
+func (Prepare) ConvcoCheck() {
```
odsod, line 26: should be `Convco`

mgtool/configuration.go
```diff
@@ -3,13 +3,17 @@ package mgtool
...
+
+	"github.com/magefile/mage/mg"
 )
 
 const (
 	AMD64 = "amd64"
 	X8664 = "x86_64"
 )
 
+type Prepare mg.Namespace
```
odsod, line 15: is this used to ignore outputting to Makefiles? maybe add as comment?

targets/mgsops/targets.go
```diff
@@ -19,15 +19,21 @@ const version = "3.7.1"
 // nolint: gochecknoglobals
 var executable string
 
+type Prepare mgtool.Prepare
+
+func (Prepare) Sops() {
+	mg.Deps(prepare)
```
odsod, line 25: could we just inline the `prepare` function here?

mgtool/configuration.go
viktorvoltaire: Yeah

targets/mgsops/targets.go
viktorvoltaire: Yep, fixed

targets/mgconvco/targets.go
viktorvoltaire: Fixed

mgtool/build.go
viktorvoltaire: The error I got, from when trying to create a symlink when one already existed was clear enough. I think we can stick with what we got

targets/mgsops/targets.go
odsod: Might as well inline all prepare functions?

odsod (approved):
:rocket:

PR #86, opened 2022-01-19
feat: replace targets with tools

viktorvoltaire:


.mage/magefile.go
```diff
@@ -41,14 +29,38 @@ func init() {
...
 	)
 	mg.SerialDeps(
-		mggo.GoModTidy,
+		GoModtidy,
 		mggitverifynodiff.GitVerifyNoDiff,
 	)
 }
+
+func FormatYaml() error {
+	return mgyamlfmt.FormatYamlCommand()
```
odsod, line 45: IMO since this is not a command it's better named `FormatYaml`

odsod: actually let's make that `FormatYAML`

.mage/magefile.go
```diff
@@ -41,14 +29,38 @@ func init() {
...
+}
+
+func GoModtidy() error {
```
odsod, line 48: `GoModTidy`? (capital `T`)

.mage/magefile.go
```diff
@@ -41,14 +29,38 @@ func init() {
...
+	return mggo.GoModTidy().Run()
+}
+
+func Goreview(ctx context.Context) error {
+	return mggoreview.Command(ctx, "-c", "1", "./...").Run()
```
odsod, line 53: really liking this pattern

example/.mage/magefile.go
```diff
@@ -38,8 +38,8 @@ func init() {
 
 func All() {
 	mg.Deps(
-		mg.F(mgconvco.ConvcoCheck, "origin/master..HEAD"),
-		mggolangcilint.GolangciLint,
+		mg.F(mgconvco.CheckCommand, "origin/master..HEAD"),
```
odsod, line 41: this should be moved to a separate function?

mgtool/exec.go
```diff
@@ -37,3 +39,14 @@ func run(cmd, dir string, stdout, stderr io.Writer, args ...string) error {
...
 }
+
+func Command(path string, args ...string) *exec.Cmd {
+	cmd := exec.Command(path)
+	cmd.Args = append(cmd.Args, args...)
+	cmd.Dir = mgpath.FromGitRoot(".")
+	cmd.Env = os.Environ()
+	cmd.Stderr = os.Stderr
+	cmd.Stdout = os.Stdout
+	cmd.Stdin = os.Stdin
```
odsod, line 50: IMO this might not be what we want

odsod: imagine multiple commands running in parallel, all trying to read from the parent process stdin

targets/mgbuf/targets.go
```diff
@@ -3,49 +3,32 @@ package mgbuf
 import (
 	"context"
 	"fmt"
+	"go.einride.tech/mage-tools/mglog"
```
odsod, line 6: formatting

targets/mggitverifynodiff/targets.go
```diff
@@ -1,17 +1,25 @@
...
+	"go.einride.tech/mage-tools/mgtool"
 )
 
 func GitVerifyNoDiff() error {
 	mglog.Logger("git-verify-no-diff").Info("verifying that git has no diff...")
-	diff, _ := sh.Output("git", "status", "--porcelain")
-	if diff != "" {
-		_ = sh.RunV("git", "diff", "--patch")
+	cmd := mgtool.Command("git", "status", "--porcelain")
+	b := &bytes.Buffer{}
```
odsod, line 14: could be simplified to `var b bytes.Buffer` and `cmd.Stdout = &b`

targets/mgwire/targets.go
```diff
@@ -1,11 +1,10 @@
...
-	"github.com/magefile/mage/sh"
-	"go.einride.tech/mage-tools/mglog"
+	"go.einride.tech/mage-tools/mgtool"
+	"os/exec"
 )
 
-func WireGenerate(path string) error {
-	mglog.Logger("wire-generate").Info("generating initializers...")
-	return sh.RunV("go", "run", "-mod=mod", "github.com/google/wire/cmd/wire", "gen", path)
+func WireGenerate(path string) *exec.Cmd {
```
odsod, line 8: IMO this might be better formulated as

```go
func Command(args ...string) *exec.Cmd {
    args = append(args, "go", "run", "-mod=mod", "github.com/google/wire/cmd/wire")
    return mgtool.Command(args)
}
```

.mage/magefile.go
```diff
@@ -41,14 +28,38 @@ func init() {
...
+		FormatYaml,
 	)
 	mg.SerialDeps(
-		mggo.GoModTidy,
+		GoModtidy,
 		mggitverifynodiff.GitVerifyNoDiff,
 	)
 }
+
+func FormatYaml() error {
```
Edholm, line 43: Are these functions added for removing the magic `// mage:import` ?

in any case, an added bonus is that you don't need to specify an argument when running only e.g. convco-check

.mage/magefile.go
Edholm: Me too :100:

.mage/magefile.go
```diff
@@ -41,14 +28,38 @@ func init() {
...
+func Goreview(ctx context.Context) error {
+	return mggoreview.Command(ctx, "-c", "1", "./...").Run()
+}
+
+func GolangciLint(ctx context.Context) error {
+	return mggolangcilint.LintCommand(ctx).Run()
+}
+
+func FormatMarkdown(ctx context.Context) error {
+	return mgmarkdownfmt.Command(ctx, "-w", ".").Run()
```
Edholm, line 60: Does this command have long flag names? :thinking: If so, I think we should switch. Makes it easier to read and get a feeling for what the command does.

mgtool/exec.go
```diff
@@ -37,3 +39,14 @@ func run(cmd, dir string, stdout, stderr io.Writer, args ...string) error {
 
 	return c.Run()
 }
+
+func Command(path string, args ...string) *exec.Cmd {
```
Edholm, line 43: I would like to see a comment/docs for this function since it is going to be so ubiquitous.

targets/mgapilinter/targets.go
```diff
@@ -17,23 +17,17 @@ import (
 const version = "1.29.3"
 
 // nolint: gochecknoglobals
-var executable string
+var commandPath string
```
Edholm, line 20: If we really want, we should be able to remove this global variable.

I'm thinking of the following:
* Specify constants for binary name, version, etc
* Add a function that builds the path based of those constants and any variables, like path to tools dir

Could be a future improvement.

targets/mggo/targets.go
```diff
@@ -2,23 +2,20 @@ package mggo
...
 )
 
-func GoTest() error {
+func GoTest() *exec.Cmd {
 	const toolName = "go"
-	mglog.Logger("go-test").Info("running Go unit tests..")
 	coverFile := filepath.Join(mgpath.Tools(), toolName, "coverage", "go-test.txt")
 	if err := os.MkdirAll(filepath.Dir(coverFile), 0o755); err != nil {
-		return err
+		panic(err)
```
Edholm, line 16: image

:joy:

Edholm (commented):
Very nice! I like the new structure

.mage/magefile.go
odsod: With the new structure for running commands centered around `exec.Cmd`, most of the public functions in the tools packages are no longer valid mage targets, so to keep the mental model simple we're thinking that we should instead avoiding having ready-made mage-targets altogether and instead write those in the `magefile.go`.

I think this will lead to a simpler mental model, since there will be a 1-1 mapping between Make targets and Go functions in `magefile.go`.

In short, yes, no more `//mage:import`!

.mage/magefile.go
odsod: I agree, but unfortunately not: https://github.com/shurcooL/markdownfmt/blob/master/main.go#L23

.mage/magefile.go
Edholm: Nice, I like it!

targets/mgapilinter/targets.go
odsod: Really like this!

.mage/magefile.go
odsod: @viktorvoltaire Minor thing, but we should rename the `targets` folder to `tools` - since it now contains tools and not targets. :smile_cat:

targets/mgapilinter/targets.go
viktorvoltaire: Actually, given that we added symlinks to all tools we download, the path is now deterministic. Should be a fairly simple change.

example/.mage/magefile.go
viktorvoltaire: Yeah I haven’t touched this one yet

targets/mgapilinter/targets.go
viktorvoltaire: However, i will look at that in a separate PR

mgtool/file.go
```diff
@@ -73,6 +79,12 @@ func FromLocal(ctx context.Context, filepath string, opts ...Opt) error {
 	if s.skipFile != "" {
 		// Check if binary already exist
 		if _, err := os.Stat(s.skipFile); err == nil {
+			if s.symlink != "" {
+				_, err := CreateSymlink(s.symlink)
+				if err != nil {
```
odsod, line 84: join the if statement with the assignment above

odsod (approved):
Just when we thought it couldn't get any better :smile:

odsod (approved):
:shipit: :rocket: :rainbow:

PR #87, opened 2022-01-20
fix: remove npm check

viktorvoltaire:
I was changing this command not to print to standard out, but then I figured this error is kinda overrated. We dont do sanity checks like this on other tools. Just let the system handle it and error out on the install command.


Edholm (approved):
:+1:

PR #94, opened 2022-01-21
Reduce API surface and rely more on *exec.Cmd API

odsod:
- feat: rename mggitverifynodiff to mggit
- feat(mggo): reduce API surface
- feat(mggolangcilint): rename LintCommand to RunCommand
- feat(mgclangformat): make FormatProtoCommand accept ...args
- feat(mggolangmigrate): remove MigrateCommand shorthand
- feat(mggooglecloudprotoscrubber): remove ScrubCommand
- chore(mgmarkdownfmt): fix naming

### DIFF
--- a/cmd/build/main.go
+++ b/cmd/build/main.go
@@ -32,7 +32,7 @@ func main() {
 	if err := os.WriteFile(makeGenGo, mgmakeGen, 0o600); err != nil {
 		panic(err)
 	}
-	cmd = mggo.GoModTidy()
+	cmd = mggo.Command("mod", "tidy")
 	cmd.Dir = mageDir
 	if err := cmd.Run(); err != nil {
 		panic(err)
--- a/.mage/magefile.go
+++ b/.mage/magefile.go
@@ -50,12 +50,12 @@ func FormatYaml() error {
 
 func GoModTidy() error {
 	mglog.Logger("go-mod-tidy").Info("tidying Go module files...")
-	return mggo.GoModTidy().Run()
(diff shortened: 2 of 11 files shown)


cmd/build/main.go
```diff
@@ -32,7 +32,7 @@ func main() {
 	if err := os.WriteFile(makeGenGo, mgmakeGen, 0o600); err != nil {
 		panic(err)
 	}
-	cmd = mggo.GoModTidy()
+	cmd = mggo.Command("mod", "tidy")
```
Edholm, line 35: `go mod tidy` seems to be sometimes used with `-v` and sometimes without. Is this intended? :)

Edholm (approved):
Nice work, just had a question regarding the `-v` flag in `go mod tidy` usage.

cmd/build/main.go
odsod: here during the init I'm thinking we want it as non-verbose as possible, while during normal repo building we want to know if modules change - WDYT?

Edholm: I agree, just wanted to know that it was intentional and not an oversight :blush:

odsod: or rather, we're not super interested in when the .mage modules are updated, but we are interested when the main/production modules are updated

viktorvoltaire (approved):
I Like this change, many of the things we first put in was a lot of guess work, where we now have a clear direction of what we want to accomplish and how and what level we want to accomplish it. 👍👍

PR #95, opened 2022-01-21
reduce api2

odsod:
- feat: reduce API surface further
- feat(mgko): reduce API surface


tools/mggo/tools.go
```diff
@@ -9,16 +9,13 @@ import (
...
-}
-
 func TestCommand() *exec.Cmd {
 	coverFile := mgpath.FromTools("go", "coverage", "go-test.txt")
 	if err := os.MkdirAll(filepath.Dir(coverFile), 0o755); err != nil {
 		panic(err)
 	}
-	return Command(
+	return mgtool.Command(
+		"go",
```
Edholm, line 18: What is your thoughts on adding the `-shuffle on` flag?

tools/mgko/tools.go
```diff
@@ -28,76 +25,19 @@ func Command(ctx context.Context, args ...string) *exec.Cmd {
 	return mgtool.Command(commandPath, args...)
 }
 
-func PublishLocalCommand(ctx context.Context) *exec.Cmd {
```
Edholm, line 31: Doesn't it make sense to have these convenience wrappers here? This is a lot of stuff to remember how to write :sweat_smile:

Edholm (approved):
Woopowoop :1st_place_medal:

tools/mgko/tools.go
odsod: They are too specific and won't generalize well to all our repos - better to have them inlined IMO

tools/mggo/tools.go
odsod: Great idea!! Adding

Edholm: Just note that it may break tests on some projects which may cause some friction :sweat_smile:

tools/mgko/tools.go
Edholm: Fair enough!

viktorvoltaire (approved):
LGTM

PR #98, opened 2022-01-21
fix(hadolint): trim space from Dockerfile list

Edholm:
hadolint is sad when given an empty space as argument.


odsod (approved):
:lady_beetle:

PR #99, opened 2022-01-21
feat: use context.Context and logr.Logger consistently

odsod:
The logger name should always be the target name, and the logger should
propagate with the context through the whole call stack.


Edholm (approved):
sweet, I like it!

PR #100, opened 2022-01-21
feat: simplify and reduce logger API surface

odsod:
Re-use more `logr` APIs.

### DIFF
--- a/cmd/build/main.go
+++ b/cmd/build/main.go
@@ -8,8 +8,9 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/go-logr/logr"
 	"github.com/magefile/mage/mage"
-	"go.einride.tech/mage-tools/mglog"
+	"go.einride.tech/mage-tools/mglogr"
 	"go.einride.tech/mage-tools/mgmake"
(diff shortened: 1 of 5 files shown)


cmd/build/main.go
```diff
@@ -19,10 +20,10 @@ import (
 var mgmakeGen []byte
 
 func main() {
-	ctx := mglog.NewContext(context.Background(), "mage-tools-build")
+	ctx := logr.NewContext(context.Background(), mglogr.New("mage-tools-build"))
```
Edholm, line 23: `ctx, cancelFunc := signal.NotifyContext(context.Background(), syscall.SIGTERM, syscall.SIGINT)
`

perhaps?

Edholm (approved):
Nice!

cmd/build/main.go
odsod: I'd rather not trap the signals until we know that we actually handle context cancellation everywhere - haven't started thinking about that yet - I think we need to use https://pkg.go.dev/os/exec#CommandContext everywhere to ensure the top-level program actually exits when the context is cancelled

PR #101, opened 2022-01-21
feat: adding terraform logo as an asset

sousandrei:
Adding terraform logo as an asset﻿


odsod:
@sousandrei Btw, the commit message should be `add terraform logo` - commit messages should use imperative form - see https://www.conventionalcommits.org/en/v1.0.0/#examples

sousandrei:
Wrong repo to host this, ma bad! :cry:

PR #102, opened 2022-01-22
feat(formatting): format Go code with gofumpt

Edholm:
By default we will format all Go code recursively from the current dir.
This implicitly relies on `GitVerifyNoDiff` to stop the execution if files were improperly formatted.

URL: https://github.com/mvdan/gofumpt

### DIFF
--- a/.mage/magefile.go
+++ b/.mage/magefile.go
@@ -38,17 +38,12 @@ func All() {
 		GoTest,
 		FormatMarkdown,
 		FormatYAML,
+		FormatGo,
 	)
 	mg.SerialDeps(
 		GoModTidy,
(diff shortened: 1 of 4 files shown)


odsod:
@Edholm Is there some way we can rely on GolangCI-Lint with --fix to handle formatting?

Edholm:
> @Edholm Is there some way we can rely on GolangCI-Lint with --fix to handle formatting?

Interesting proposition, I'll experiment with it!

Edholm:
Hm I see now that we default to use `--fix` in golangci-lint and have gofumpt enabled :grinning: 

Closing PR as already implemented!

PR #106, opened 2022-01-22
feat(mgbalenacli): add balena-cli tool

Edholm:
Allows you to execute arbitrary commands using the 'balena' tool.

Also added a `Output` and `MustOutput` to mgtool. These functions makes it very easy to run git commands and retrieve the output immediately. Easy to run e.g. "git describe --tags" or similar and store the output in a string.

### DIFF
--- a/mgtool/exec.go
+++ b/mgtool/exec.go
@@ -2,14 +2,17 @@ package mgtool
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"os/exec"
+	"strings"
 
 	"go.einride.tech/mage-tools/mgpath"
 )
 
 // Command should be used when returning exec.Cmd from tools to set opinionated standard fields.
 func Command(_ context.Context, path string, args ...string) *exec.Cmd {
+	// TODO: use exec.CommandContext when we have determined there are no side-effects.
 	cmd := exec.Command(path)
 	cmd.Args = append(cmd.Args, args...)
 	cmd.Dir = mgpath.FromGitRoot(".")
@@ -18,4 +21,15 @@ func Command(_ context.Context, path string, args ...string) *exec.Cmd {
 	cmd.Stderr = os.Stderr
 	cmd.Stdout = os.Stdout
 	return cmd
+}
+
+// Output runs the given command, and returns all output from stdout in a neatly, trimmed manner,
+// panicking if an error occurs.
+func Output(cmd *exec.Cmd) string {
+	cmd.Stdout = nil
+	output, err := cmd.Output()
+	if err != nil {
+		panic(fmt.Sprintf("%s failed: %v", cmd.Path, err))
+	}
+	return strings.TrimSpace(string(output))
 }
--- a/tools/mgbalenacli/tools.go
+++ b/tools/mgbalenacli/tools.go
@@ -1,0 +1,86 @@
+package mgbalenacli
+
+import (
+	"context"
+	"fmt"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"strings"
+
+	"github.com/magefile/mage/mg"
+	"go.einride.tech/mage-tools/mgpath"
+	"go.einride.tech/mage-tools/mgtool"
+)
+
+const (
+	toolName   = "balena-cli"
+	binaryName = "balena"
+	version    = "v13.1.11"
+)
+
+func Command(ctx context.Context, args ...string) *exec.Cmd {
+	mg.CtxDeps(ctx, Prepare.Balena)
+	return mgtool.Command(ctx, mgpath.FromBins(binaryName), args...)
+}
+
+func Whoami(ctx context.Context) (WhoamiInfo, error) {
+	cmd := Command(ctx, "whoami")
+	cmd.Stdout = nil
+	output, err := cmd.Output()
+	if err != nil {
+		return WhoamiInfo{}, fmt.Errorf("balena whoami failed: %w", err)
+	}
+	lines := strings.Split(strings.TrimSpace(string(output)), "\n")[1:]
+	if len(lines) != 3 {
+		return WhoamiInfo{}, fmt.Errorf("unexpected output from Balena: %q", output)
+	}
+
+	// Example output we need to trim.
+	// == ACCOUNT INFORMATION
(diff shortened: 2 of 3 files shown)


tools/mggit/tools.go
```diff
@@ -25,21 +25,22 @@ func VerifyNoDiff(ctx context.Context) error {
...
 
-func Version(ctx context.Context) string {
-	cmd := mgtool.Command(ctx, "git", "rev-parse", "--verify", "HEAD")
-	var b bytes.Buffer
-	cmd.Stdout = &b
-	if err := cmd.Run(); err != nil {
-		panic(err)
+// RunCommand runs git with the given arguments, and returning the output as a string.
+// It will panic if an error occurs.
+func RunCommand(ctx context.Context, args ...string) string {
```
viktorvoltaire, line 30: Feels like this one deviating a bit from all the other “Command” apis we have, where they generally returns a exec.Cmd . However, it feels like this package is more of a helper one, then a tool. Should it be moved some place else? Given we are breaking the pattern with it. For it to remain here it should really just be a Command that returns git as a exec.Cmd

viktorvoltaire: To add, not saying what you added here is bad 😅 I just feel we should write down the idea we have with the general design, so it’s easier to reason about stuff like this :) fyi @odsod 
Let’s add some design philosophies for the repo?

Edholm: Yes, I thought about that too and felt this one warranted a side-step from the convention. With that said it should probably be renamed to "Run" if we decide to keep it.

The reason I added this function without the `exec.Command` is that you can do this:
```go
	var (
		gitTag    = mggit.RunCommand(ctx, "describe", "--tags")
		gitCommit = mggit.RunCommand(ctx, "log", "-1", "--pretty=format:%H")
		gitAuthor = mggit.RunCommand(ctx, "show", "-s", "--format=%an <%ae>", gitCommit)
		gitBranch = mggit.RunCommand(ctx, "rev-parse", "--abbrev-ref", "HEAD")
	)

```
which would be major hassle with the command pattern used elsewhere.

odsod: This looks more in line with the `GetOutput` function of the exec package - maybe we can find a name that points more in that direction?

odsod: I think we should rather add a helper function that takes an exec.Command and returns the stdout output and panics if there's an error. That can be combined with the conventional Command functions from all the packages?

Edholm: Yes, I was thinking you could have two conventions, one of returning an exec.Command and one that is a convenience layer on top of that. 

Maybe straight up a `MustOutput` ? or simply `Output`?

Edholm: hm good point, then it could be re-used elsewhere. I'll see if I can try something out

mgtool/exec.go
```diff
@@ -19,3 +22,24 @@ func Command(_ context.Context, path string, args ...string) *exec.Cmd {
 	cmd.Stdout = os.Stdout
 	return cmd
 }
+
+// Output runs the given command, and returns all output from stdout in a neatly, trimmed manner,
+// or an error if the command failed.
+func Output(cmd *exec.Cmd) (string, error) {
```
viktorvoltaire, line 28: Should it be OutputCommand ?

odsod, line 28: Let's call it GetOutput?

Edholm: I simply followed the naming used by the command struct :)

odsod: My bad! :speak_no_evil: :smile:

Edholm: Had a talk with Oscar and we decided to only keep the "happy-path" function. So I removed this one.

Edholm: Function is now removed :)

tools/mgbalenacli/tools.go
```diff
@@ -0,0 +1,86 @@
...
+
+func Command(ctx context.Context, args ...string) *exec.Cmd {
+	mg.CtxDeps(ctx, Prepare.Balena)
+	return mgtool.Command(ctx, mgpath.FromBins(binaryName), args...)
+}
+
+func Whoami(ctx context.Context) (WhoamiInfo, error) {
+	cmd := Command(ctx, "whoami")
+	cmd.Stdout = nil
+	output, err := cmd.Output()
```
viktorvoltaire, line 30: Shouldnt this use the new `Output` function? :)

odsod (approved):
:rocket: :rainbow: :taco:

tools/mggit/tools.go
```diff
@@ -4,19 +4,18 @@ import (
...
 
 	"go.einride.tech/mage-tools/mgtool"
 )
 
 func VerifyNoDiff(ctx context.Context) error {
 	cmd := mgtool.Command(ctx, "git", "status", "--porcelain")
-	var b bytes.Buffer
-	cmd.Stdout = &b
+	var status bytes.Buffer
+	cmd.Stdout = &status
```
viktorvoltaire, line 14: Should use `mgtool.Output` ?

Edholm: I didn't want to change the behaviour of the function (it returns an error and doesn't panic). 

What do you think? We can change in a follow-up PR if so.

tools/mgbalenacli/tools.go
Edholm: Nope, this function needs to return an error so the caller can intercept failure (i.e. when not logged in for example)

tools/mggit/tools.go
viktorvoltaire: I was more meaning the part where we are setting stdout and then checking that status.String is empty. That part can easly just be changed to mgtool.Output? given that we have changed this anyway

viktorvoltaire: Oh yeah, i see what you mean now. Sure lets keep it

tools/mgbalenacli/tools.go
viktorvoltaire: :+1:

viktorvoltaire (approved):
:+1:

PR #107, opened 2022-01-22
feat: refactor how we generate binary and makefiles

viktorvoltaire:
This PR speeds up the build process and cuts out a lot of unnecessary steps that we previously did to generate both binary and makefiles. It also removes the need to have special mage tags in the go files.

One other big thing, is that I decided to not implement the functionality that can generate targets via imports. This is something we had support for with the old way of doing things. But it was hard to understand if you weren’t fully aware of how everything works, so this essentially makes it so that everything needs to be declared, which is something we have moved towards anyway


.mage/makefile.go
```diff
@@ -0,0 +1,22 @@
...
+import (
+	"context"
+	"github.com/go-logr/logr"
+	"go.einride.tech/mage-tools/mglogr"
+
+	"go.einride.tech/mage-tools/mgmake"
+	"go.einride.tech/mage-tools/mgpath"
+)
+
+func main() {
```
viktorvoltaire, line 12: This would be easy to code generate with the init command

mgmake/generate.go
```diff
@@ -69,20 +73,39 @@ func GenerateMakefiles(mks ...Makefile) {
...
+func GenMakefiles(ctx context.Context) {
 	if len(makefiles) == 0 {
-		return fmt.Errorf("no makefiles to generate, see https://github.com/einride/mage-tools#readme for more info")
+		panic("no makefiles to generate, see https://github.com/einride/mage-tools#readme for more info")
 	}
-	executable = exec
-	targets, err := listTargets(ctx)
+	mageFilesCmd := mgtool.Command(ctx, "go", "list", "-e", "-f", `{{join .GoFiles "\n"}}`)
+	var b bytes.Buffer
+	mageFilesCmd.Stdout = &b
```
Edholm, line 87: Just an FYI that this can be replaced with `.Output()` which is run and output in a single command. Doesn't work when stdout is non-nil though. Any reason we're using this pattern instead? :thinking:

Edholm (approved):
Looks good, nice work!

mgmake/generate.go
viktorvoltaire: Actually, i just changed this to `filepath.WalkDir`, was looking a bit to much at the mage implementation when i wrote it :)

.mage/makefile.go
```diff
@@ -0,0 +1,15 @@
+package main
+
+import (
+	"go.einride.tech/mage-tools/mgmake"
+	"go.einride.tech/mage-tools/mgpath"
+)
+
+func main() {
+	mgmake.GenerateMakefiles(
```
viktorvoltaire, line 9: I wish we could solve this in some other way, the only reason I had to go for this approach is because we need to set the global variable which this method sets in order to know which makefikes to generate.

open to suggestions, can we solve this in some other way? 🤔

also maybe we should sanity check this file, since it’s ignored later when we compile, should we make sure there are no other methods or something?

Makefile
```diff
@@ -14,36 +14,36 @@ clean-mage-tools:
 
 .PHONY: all
 all: $(magefile)
-	@$(magefile) all
```
odsod, line 17: Nice! This makes so much sense

cmd/build/main.go
```diff
@@ -3,49 +3,33 @@ package main
...
-	cmd.Dir = mageDir
-	cmd.Stdout = io.Discard
-	if err := cmd.Run(); err != nil {
-		panic(err)
-	}
 	if err := os.WriteFile(makeGenGo, mgmakeGen, 0o600); err != nil {
 		panic(err)
 	}
-	cmd = mgtool.Command(ctx, "go", "mod", "tidy")
+	defer os.Remove(makeGenGo)
```
odsod, line 26: Might want to defer an inline function which checks the error, in case removal fails for some reason

odsod (approved):
Nice!

PR #109, opened 2022-01-24
feat: change generated magefile to use master instead of main

alethenorio:
Github (and I think the industry at large?) is still using master as the default branch name for new repos so it makes more sense to generate 
magefile for master rather than main until this has been changed.

### DIFF
--- a/cmd/init/example/.mage/magefile.go
+++ b/cmd/init/example/.mage/magefile.go
@@ -81,7 +81,7 @@ func FormatMarkdown(ctx context.Context) error {
 func ConvcoCheck(ctx context.Context) error {
 	ctx = logr.NewContext(ctx, mglogr.New("convco-check"))
 	logr.FromContextOrDiscard(ctx).Info("checking git commits...")
-	return mgconvco.Command(ctx, "check", "origin/main..HEAD").Run()
+	return mgconvco.Command(ctx, "check", "origin/master..HEAD").Run()
 }
 
 func GitVerifyNoDiff(ctx context.Context) error {


cmd/init/example/.mage/magefile.go
```diff
@@ -81,7 +81,7 @@ func FormatMarkdown(ctx context.Context) error {
 func ConvcoCheck(ctx context.Context) error {
 	ctx = logr.NewContext(ctx, mglogr.New("convco-check"))
 	logr.FromContextOrDiscard(ctx).Info("checking git commits...")
-	return mgconvco.Command(ctx, "check", "origin/main..HEAD").Run()
+	return mgconvco.Command(ctx, "check", "origin/master..HEAD").Run()
```
thall, line 84: We have repos with both main and master, cant we instead make two different specialized functions or a more general one which takes a revision?
```go
func ConvcoCheckMasterHead(ctx context.Context) error
func ConvcoCheckMainHead(ctx context.Context) error
func ConvcoCheck(ctx context.Context, revision string) error
```

alethenorio: This is for the generated target which ends up in the "all" target. It is a little difficult to run both and it made sense to me to default to whatever the industry defaults to right now. Other than that it is a simple change for the user to fix the branch from "master" to "main" which makes sense since you will likely be adapting the magefile a little yo each repo anyway.

What do you think?

thall: Ah I see what you mean. Would be nice if we could check if we are using master och main branch before generating the target.
I'm fine with either way, just stick to one.

alethenorio: @odsod has tried it. It is unfortunately not so simple.

odsod (approved):
All good!

PR #110, opened 2022-01-24
feat: import mage

odsod:
This commit forklifts the essential parts of mage into this repository,
and removes all non-essential parts.

The unmodified code has LICENSE files and license headers applied to
show what has been forklifted.

The near-term goal is to keep simplifying and start making a few
additions, such as automatically including a logr.Logger in the context
of all targets.


Edholm (approved):
Woop woop

PR #111, opened 2022-01-25
feat: use native git-go for deriving git root

alethenorio:
Since git has a native Go implementation, we make use of it to derive the git root instead.
See https://git-scm.com/book/en/v2/Appendix-B%3A-Embedding-Git-in-your-Applications-go-git for more details


.mage/go.mod
```diff
@@ -9,8 +9,25 @@ require (
 )
 
 require (
+	github.com/Microsoft/go-winio v0.4.16 // indirect
```
odsod, line 12: TBH I think this pulls in too many dependencies

alethenorio: Yeah it does bring more dependencies than one would hope. My thinking for going forward with it was mostly that we would no longer rely on having to call "exec.Command" and that it is recommended by the "official got documentation" as linked above but if we feel that the dependency surface is an issue I am OK with that

odsod: Yeah I see where you're going but this was an unexpectedly heavy dependency - since we will use this repo quite widely I was even thinking we would try and remove a few existing dependencies (google/uuid and maybe even strcase) just to keep this repo quite stable over time.

alethenorio: Fair. I took a second pass on the dependencies and they are not that many once you look at them (We would likely depend on the same if we were to do something similar) but git is likely not the biggest concern for potential source of issues due to local missing tooling

alethenorio:
Closing due to the large dependency surface of go-git in favor of keeping stability on the project

PR #112, opened 2022-01-25
feat: use module-aware mode when building Go from local mod file

alethenorio:
Often the purpose for wanting to build a Go binary from a remote package using the local go.mod file is so that we build the binary based on the same version as the local repo depends on however we don't want the local go.mod to include dependencies only needed to build that binary (which may be different than the "library" part of the remote repo) so we use the version specified in the local go.mod file but build the binary in module-aware mode so it brings its own dependencies


odsod (approved):
Great thinking!

PR #113, opened 2022-01-25
fix: use os.Lstat instead of os.Stat for symlinks

alethenorio:
`os.Stat` follows symlinks when the intention here is to check whether the symlink exists and remove it if it does. For that we need to use os.Lstat which does not follow symlinks


odsod (approved):
SGTM

PR #122, opened 2022-01-27
feat: introduce terraform tool

alethenorio:
It is unclear to me whether the `PrepareCommand(context.Context)` function fulfills any purpuse that is not clear in the API (Does not seem to be used anywhere so why is it public? Do we intend for users to be able to use it directly?) so I kept it here for consistency with the other packages.

### DIFF
--- a/tools/sgterraform/tools.go
+++ b/tools/sgterraform/tools.go
@@ -1,0 +1,47 @@
+package sgterraform
+
+import (
+	"context"
+	"fmt"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+
+	"go.einride.tech/sage/sg"
+	"go.einride.tech/sage/sgtool"
+)
+
+const (
+	version    = "1.1.4"
+	binaryName = "terraform"
+)
+
+func Command(ctx context.Context, args ...string) *exec.Cmd {
+	return sg.Command(ctx, sg.FromBinDir(binaryName), args...)
+}
+
+func PrepareCommand(ctx context.Context) error {
+	hostOS := runtime.GOOS
+	hostArch := runtime.GOARCH
+	binaryDir := sg.FromBinDir(binaryName)
+	binary := filepath.Join(binaryDir, binaryName)
+	terraform := fmt.Sprintf("terraform_%s_%s_%s", version, hostOS, hostArch)
+	binURL := fmt.Sprintf(
+		"https://releases.hashicorp.com/terraform/%s/%s.zip",
+		version,
+		terraform,
+	)
+	if err := sgtool.FromRemote(
+		ctx,
+		binURL,
+		sgtool.WithDestinationDir(binaryDir),
+		sgtool.WithUnzip(),
+		sgtool.WithRenameFile(fmt.Sprintf("%s/terraform", terraform), binaryName),
+		sgtool.WithSkipIfFileExists(binary),
+		sgtool.WithSymlink(binary),
+	); err != nil {
+		return fmt.Errorf("unable to download %s: %w", binaryName, err)
+	}
+
+	return nil
+}


Edholm:
Just an FYI that we have terraform in the internal mage-tools repo, but I think the idea is to migrate it here sooner or later.

tools/sgterraform/tools.go
```diff
@@ -0,0 +1,58 @@
...
+	return sg.Command(ctx, filepath.Join(binDir, binaryName), args...)
+}
+
+func PrepareCommand(ctx context.Context) error {
+	toolDir := sg.FromToolsDir(binaryName)
+	binDir := filepath.Join(toolDir, version, "bin")
+	return prepareCommand(ctx, version, binDir, binaryName)(ctx)
+}
+
+func prepareCommand(ctx context.Context, version, binaryDir, binaryName string) func(context.Context) error {
```
odsod, line 32: This looks like it can be inlined into PrepareCommand

odsod:
@alethenorio The PrepareCommand function ensures that, when it's used as a dependency with `sg.Deps` it will only be run once (i.e. the terraform binary will only be downloaded once) even when there are multiple uses of the tool.

odsod:
@Edholm AFAIK there's a bunch of internal "recipes" and helper functions in the internal repo that we might not want to move to the public repo. But having the terraform CLI tool available here will be great!

tools/sgterraform/tools.go
odsod: @viktorvoltaire Was it you who wrote this initially? I'm guessing this is from when we were still making version configurable - now that we prefer constant versions I think we can remove the parameters and inline this into the public PrepareCommand function.

odsod: @alethenorio Having the PrepareCommand function public in all packages is meant to enable downloading the tool without using it from the Sagefile itself, for example when it's needed in the PATH by another tool, or something of the sort. We have encountered a few use cases like that while rolling out to various repos.

alethenorio: The main purpose of this function is to avoid having to go through a global variable for setting up the command path. Inlining it into the "Public PrepareCommand" would sacrifice that for no real gain

odsod: Can we solve that that with a private function that calculates the command path on the fly instead?

The two different prepare functions will be counted as different deps and can cause terraform to be downloaded twice.

viktorvoltaire: Some context:
Version is supported in our internal library because it was mentioned that Terraform is one of those tools where the consumer might want more control, given that terraform has a tendency to change things that might be unwanted in between versions.

You should be able to just set binaryDir and binaryName as constants, then utilize the fact that `WithSymlink(binary)` gives a known path, which would be `.sage/bin/terraform`

alethenorio: Do we see a use for having the Prepare function public? Cause otherwise we could skip it. What do you think?

If the user for some reason decides to use both Prepare and Command (Which I think will be rare), we won't get terraform downloaded twice since the `sgtool.WithSkipIfFileExists(binary)` will prevent that but I agree it is not ideal.

Can give this some further thought

alethenorio: I didn't take into account supporting version anywhere here. It was kept simple on purpose since it is helpful to have the binary but less so a bunch of targets which are more internal than not.

Here's how we have implemented in the vehicle-service as an example (which I would hope to move the binary and command parts out to here)

https://github.com/einride/vehicle-service/blob/master/.sage/terraform.go

viktorvoltaire: I feel like we can just default, like we have done with all other tools and have it as Public, we can add it to the documentation to make it clear why we have it like this. Makes it easy when adding new tools to just follow one pattern. 

I do feel like we should hardcode version, and if the consumer wants a different one, they can go a head and just inline the prepare command with their version in their repository

alethenorio: Yeah I agree on that.

alethenorio: I realized that it was just possible to depend on the Public PrepareCommand in the Command function so the problem with potential duplicate of attempting to download the tool is fixed :sunglasses:

alethenorio: Alright. Took another pass at it. Think this is pretty good now. Only reusing a small function for the bin path

odsod (approved):
Sweet

tools/sgterraform/tools.go
odsod: Nice one!

alethenorio:
> Just an FYI that we have terraform in the internal mage-tools repo, but I think the idea is to migrate it here sooner or later.

Yeah I didn't realize this but I also think it is a good idea to have the tool here anyway

tools/sgterraform/tools.go
```diff
@@ -0,0 +1,52 @@
...
+	"go.einride.tech/sage/sgtool"
+)
+
+const (
+	version    = "1.1.4"
+	binaryName = "terraform"
+)
+
+func Command(ctx context.Context, args ...string) *exec.Cmd {
+	return sg.Command(ctx, filepath.Join(binDir(), binaryName), args...)
```
viktorvoltaire, line 20: Bindir can be replaced with `sg.FromBinDir(binaryName)` and with that you can also remove the binDir function

viktorvoltaire (changes requested):

tools/sgterraform/tools.go
alethenorio: Hurray for collaboration. Fixed. Too bad it can't be used in a constant

viktorvoltaire (approved):
Awesome! :partying_face: :tada:

PR #124, opened 2022-01-27
feat: migrate back from cmd to git root main

viktorvoltaire:


Edholm (approved):
Is this due to ease-of-use or ?

odsod:
@Edholm Yes, it creates a bit more "productified" CLI interface when it's one CLI with subcommands - and as an added bonus we can move the example(s) up to the top level!

Makefile
```diff
@@ -6,7 +6,7 @@
 sagefile := .sage/tools/bin/sagefile
 
 $(sagefile): .sage/go.mod $(shell find .sage/.. -type f -name '*.go')
-	@cd .sage && go run ../cmd/build
+	@cd .sage && go run ../main.go run
```
odsod, line 9: did we rename it to `run` instead of `build`?

odsod (approved):

odsod (approved):
Wow!

PR #126, opened 2022-01-28
golangci lint run

odsod:
GolangCI-Lint only runs for the Go module of the current working
directory, which is why we've seen some issues with the .sage directory
not being automatically formatted or checked by the linter.


tools/sggolangcilint/tools.go
```diff
@@ -14,30 +15,57 @@ import (
...
-		if err := os.WriteFile(configPath, []byte(defaultConfig), 0o600); err != nil {
-			panic(err)
+// Run GolangCI-Lint in every Go module from the root of the current git repo.
+func Run(ctx context.Context) error {
+	defaultConfigPath := sg.FromToolsDir("golangci-lint", ".golangci.yml")
+	if err := os.MkdirAll(filepath.Dir(defaultConfigPath), 0o755); err != nil {
+		return err
+	}
+	if err := os.WriteFile(defaultConfigPath, defaultConfig, 0o600); err != nil {
+		panic(err)
```
Edholm, line 38: I vote return err here since we do elsewhere.

Edholm (approved):
Neat!

tools/sggolangcilint/tools.go
odsod: Ooops! Thanks!

tools/sggolangcilint/golangci.yml
```diff
@@ -1,7 +1,9 @@
 run:
   timeout: 10m
-  skip-dirs:
```
viktorvoltaire, line 3: Can this cause a problem for consumers? Should they be able to set this without having the entire config?

Edholm (approved):

tools/sggolangcilint/golangci.yml
odsod: Very good point! I'm updating so that the user can pass `...args string` to run and use flags to override specific stuff

viktorvoltaire: 👍

odsod: Updated!

viktorvoltaire (approved):
Really nice!

PR #127, opened 2022-01-28
feat(sgtool): remove unused methods

odsod:


Edholm (approved):
:broom:

viktorvoltaire:
Just because something isn’t used in this repo, it doesn’t mean it isn’t used by others. I think both of these are used outside this repo

PR #130, opened 2022-01-28
feat: remove external dependencies

odsod:


sg/logger.go
```diff
@@ -1,56 +1,32 @@
...
 }
 
-func (s *logSink) WithName(name string) logr.LogSink {
-	s.name = name
-	return s
+// Logger returns the log.Logger attached to the provided context.
+func Logger(ctx context.Context) *log.Logger {
+	logger := ctx.Value(loggerContextKey{}).(*log.Logger)
+	if logger == nil {
+		panic("context is missing a Logger")
```
gust1n, line 29: Would it make sense to provide a default named logger if it's missing, instead of panicking?

odsod: Hmm! Good point!

odsod: Updated

sg/logger.go
```diff
@@ -1,56 +1,31 @@
...
-	panic("implement me")
+// WithLogger attaches a log.Logger to the provided context.
+func WithLogger(ctx context.Context, logger *log.Logger) context.Context {
+	return context.WithValue(ctx, loggerContextKey{}, logger)
 }
 
-func (s *logSink) WithName(name string) logr.LogSink {
-	s.name = name
-	return s
+// Logger returns the log.Logger attached to the provided context.
```
gust1n, line 25: Then I also think the comment should make that clear (a default logger will be returned if the context doesn't contain one)

odsod: Updated

gust1n:
LGTM!

viktorvoltaire (approved):
Love it, let’s go! 🥳

PR #133, opened 2022-01-28
fix(terraform): prepare command on terraform tool

sousandrei:
Terraform was not preparing the command﻿

PR #134, opened 2022-01-28
fix: cleanup the clean target

odsod:


sg/makefile.go
```diff
@@ -42,7 +42,7 @@ func generateMakefile(g *codegen.File, pkg *doc.Package, mk Makefile, mks ...Mak
 	g.P()
 	g.P(".PHONY: clean-sage")
 	g.P("clean-sage:")
-	g.P("\t@git clean -fdx ", filepath.Join(includePath, ToolsDir))
+	g.P("\t@git clean -fdx ", includePath)
```
Edholm, line 45: this now removes the bin dir also right? 

Then @sousandrei can remove his fix for that in #133

Edholm (approved):
:+1:

sg/makefile.go
sousandrei: opsie

odsod: rebased :smile_cat:

Edholm (approved):

PR #135, opened 2022-01-28
fix: revert "fix: cleanup the clean target"

odsod:
This reverts commit 97aafa06a99fab1e849bb300c411a7c463339e65.

As @viktorvoltaire pointed out, if we clean at the `.mage` level, we
will "clean" away any work-in-progress sagefiles - that would not be
good!


Edholm (approved):
Good point! :+1:

viktorvoltaire:
👍

PR #138, opened 2022-01-28
add tf pr comment command

viktorvoltaire:
This would be called with something like:

```golang
func (Terraform) CommentOnPullRequestWithPlan(ctx context.Context, pr, project string) error {
	cmd = sgterraform.CommentOnPullRequestWithPlan(ctx, pr, project, sgpath.FromGitRoot(filepath.Join(terraformDir, "terraform.plan"))
	cmd.Dir = sgpath.FromGitRoot(terraformDir)
	return cmd.Run()
}
```

### DIFF
--- a/tools/sgterraform/tools.go
+++ b/tools/sgterraform/tools.go
@@ -9,6 +9,7 @@ import (
 
 	"go.einride.tech/sage/sg"
 	"go.einride.tech/sage/sgtool"
+	"go.einride.tech/sage/tools/sgghcomment"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


tools/sgterraform/tools.go
```diff
@@ -21,6 +22,30 @@ func Command(ctx context.Context, args ...string) *exec.Cmd {
 	return sg.Command(ctx, sg.FromBinDir(binaryName), args...)
 }
 
+func CommentPRWithTFPlan(ctx context.Context, prNumber, environment, tfPlan string) *exec.Cmd {
```
odsod, line 25: Can remove TF to avoid repetition - how about `CommentOnPullRequestWithPlan`?

viktorvoltaire: Yes!

viktorvoltaire:
@odsod what do you think about my latest push, with this the function in the sagefile would look like:
```golang
func (Terraform) CommentOnPullRequestWithPlan(ctx context.Context, pr, project string) error {
	cmd = sgterraform.CommentOnPullRequestWithPlan(ctx, pr, project, sgpath.FromGitRoot(terraformDir, "terraform.plan"))
	cmd.Dir = sgpath.FromGitRoot(terraformDir)
	return cmd.Run()
}
```

odsod (approved):
I feel like it definitely adds some value - but do we want to standardize this so soon? All for it if you think so!

viktorvoltaire:
> I feel like it definitely adds some value - but do we want to standardize this so soon? All for it if you think so!

It does look exactly like this across the board, and the function itself has a very narrow use case and there is basically no room for flexibility. 

I feel confident, but by saying so I will definitely eat this up in the future 😀🧙‍♀️

odsod (approved):
I'm sold :smile_cat:

PR #139, opened 2022-01-28
feat: pipe stdout and err via logger

viktorvoltaire:
Still not a 100% sure what I feel about this one, it does make it easier to understand which command the stdout comes from when running stuff in parallel. But it also just feels not 100% natural when seing the oudput in my terminal :P

### DIFF
--- a/sg/logger.go
+++ b/sg/logger.go
@@ -29,8 +29,8 @@ func WithLogger(ctx context.Context, logger *log.Logger) context.Context {
 
 // Logger returns the log.Logger attached to ctx, or a default logger.
 func Logger(ctx context.Context) *log.Logger {
-	if logger := ctx.Value(loggerContextKey{}).(*log.Logger); logger != nil {
-		return logger
+	if value := ctx.Value(loggerContextKey{}); value != nil {
+		return value.(*log.Logger)
 	}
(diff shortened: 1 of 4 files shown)


sg/logger.go
```diff
@@ -24,8 +24,10 @@ func WithLogger(ctx context.Context, logger *log.Logger) context.Context {
 
 // Logger returns the log.Logger attached to ctx, or a default logger.
 func Logger(ctx context.Context) *log.Logger {
-	if logger := ctx.Value(loggerContextKey{}).(*log.Logger); logger != nil {
-		return logger
+	if ok := ctx.Value(loggerContextKey{}); ok != nil {
+		if logger := ok.(*log.Logger); logger != nil {
```
odsod, line 28: This should not be necessary since it's our own private functions that assign the logger value - it's impossible to get any other type than `*log.Logger` out.

viktorvoltaire: That’s what I thought, however i get a panic without this 🤷‍♀️

viktorvoltaire: But I guess I could try and find out why

odsod: hmm yes, could you post a stack trace?

odsod (approved):

PR #142, opened 2022-01-29
feat: introduce build/ dir

odsod:
For build artifacts, such as test coverage reports and similar outputs
from running a build.

### DIFF
--- a/sg/path.go
+++ b/sg/path.go
@@ -9,10 +9,11 @@ import (
 )
 
 const (
-	SageDir        = ".sage"
-	ToolsDir       = "tools"
-	BinDir         = "bin"
-	SageFileBinary = "sagefile"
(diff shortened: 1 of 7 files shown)


sg/path.go
```diff
@@ -15,6 +15,10 @@ const (
 	SageFileBinary = "sagefile"
 )
 
+const (
+	buildDir = "build"
```
viktorvoltaire, line 19: Is this in its own group because it’s private?

viktorvoltaire (approved):

sg/path.go
odsod: yes, but now I pushed an update where we make them all private!

viktorvoltaire: Sweet!

PR #144, opened 2022-01-30
fix: ensure unique IDs for functions with different args

odsod:
Regression from the mage-tools migration, fixed by introducing a unique
ID different from the display name.

### DIFF
--- a/sg/fn.go
+++ b/sg/fn.go
@@ -2,6 +2,7 @@ package sg
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"reflect"
 	"runtime"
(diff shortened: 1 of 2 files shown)


Jassob (approved):
Looks good to me! :+1: :star:

sg/fn.go
```diff
@@ -70,8 +74,14 @@ func newFn(f interface{}, args ...interface{}) (Function, error) {
...
 	}
+	argsID, err := json.Marshal(args)
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate JSON name for args: %w", err)
+	}
+	name := runtime.FuncForPC(reflect.ValueOf(f).Pointer()).Name()
 	return fn{
-		name: runtime.FuncForPC(reflect.ValueOf(f).Pointer()).Name(),
+		name: name,
+		id:   name + "(" + string(argsID) + ")",
```
alethenorio, line 84: Is this the name of the arguments or the actual values passed? Caused the former won't solve the problem right?

alethenorio (approved):

sg/fn.go
odsod: actual values passed, so this should do the trick

viktorvoltaire:
Nice @odsod 👌

PR #145, opened 2022-01-31
feat(sggit): use short hash when deriving version

alethenorio:
The short hash is enough to derive the commit

### DIFF
--- a/tools/sggit/tools.go
+++ b/tools/sggit/tools.go
@@ -26,7 +26,7 @@ func VerifyNoDiff(ctx context.Context) error {
 
 func Version(ctx context.Context) string {
 	revision := sg.Output(
-		sg.Command(ctx, "git", "rev-parse", "--verify", "HEAD"),
+		sg.Command(ctx, "git", "rev-parse", "--verify", "--short", "HEAD"),
 	)
 	diff := sg.Output(
 		sg.Command(ctx, "git", "status", "--porcelain"),


tools/sggit/tools.go
```diff
@@ -26,7 +26,7 @@ func VerifyNoDiff(ctx context.Context) error {
 
 func Version(ctx context.Context) string {
 	revision := sg.Output(
-		sg.Command(ctx, "git", "rev-parse", "--verify", "HEAD"),
+		sg.Command(ctx, "git", "rev-parse", "--verify", "--short", "HEAD"),
```
alethenorio, line 29: Do we want to have a bool argument to optionally return the long version?

thall: hijacking this a bit, in the driver-app we have start using `git describe --tags --long --always` to get version which includes the tag i.e  `v1.104.0-0-deadbeef`, `v1.104.0-1-g4abde7660`, it requires that the tags are available otherwise it fallback to only short commit hash. Could be nice to have a method for that as well?

Edholm: If one project have special needs for the version I think they could just re-implement it. Let's cater for the general case

Edholm (approved):
Looks good :+1:

tools/sggit/tools.go
alethenorio: I agree with this though I have to admit the format is not bad for the one you described Niclas.

PR #146, opened 2022-01-31
fix: correctly use `fatalf` instead of `fatal`

Edholm:
in the generated code. Produces correct error output in certain scenarios.


viktorvoltaire (approved):
Nice!

PR #147, opened 2022-01-31
fix(init): expected number of arguments

Edholm:
the first expected argument is the context, and that isn't provided
as an application argument hence the minus one.


viktorvoltaire (approved):
Perfect!

PR #152, opened 2022-02-01
feat: add target for building sage

Holmus:
When migrating to sage in a repository yesterday I found that quite a lot of time I would go back into the sage directory to regenerate my Makefile. Adding a target for doing this could be useful.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,8 +5,12 @@
 
 sagefile := .sage/bin/sagefile
 
-$(sagefile): .sage/go.mod $(shell find .sage/.. -type f -name '*.go')
+.PHONY: $(sagefile)
+$(sagefile):
 	@cd .sage && go mod tidy && go run .
+
+.PHONY: sage
+sage: $(sagefile)
 
 .PHONY: clean-sage
 clean-sage:
--- a/sg/generate.go
+++ b/sg/generate.go
@@ -68,7 +68,7 @@ func GenerateMakefiles(mks ...Makefile) {
 		mk := codegen.NewMakefile(codegen.FileConfig{
 			GeneratedBy: "go.einride.tech/sage",
 		})
-		if err := generateMakefile(ctx, mk, pkg, v, mks...); err != nil {
+		if err := generateMakefile(mk, pkg, v, mks...); err != nil {
 			panic(err)
 		}
 		if err := os.WriteFile(v.Path, mk.RawContent(), 0o600); err != nil {
--- a/sg/makefile.go
+++ b/sg/makefile.go
@@ -1,7 +1,6 @@
 package sg
 
 import (
-	"context"
 	"fmt"
 	"go/ast"
 	"go/doc"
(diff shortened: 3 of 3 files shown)


Edholm:
What is the reason for addition? :thinking: 

Usually a `touch .sage/go.mod && make` should cover this use-case

Holmus:
> What is the reason for the addition? thinking
> 
> Usually a `touch .sage/go.mod && make` should cover this use-case

I find using a target for this easier.

Edholm (commented):
> > What is the reason for the addition? thinking
> > Usually a `touch .sage/go.mod && make` should cover this use-case
> 
> I find using a target for this easier.

I have nothing against it. Have you tried it in your repo? Does it work the way you expect it to?

I'll leave it to @odsod or @viktorvoltaire to approve though

Holmus:
> > > What is the reason for the addition? thinking
> > > Usually a `touch .sage/go.mod && make` should cover this use-case
> > 
> > 
> > I find using a target for this easier.
> 
> I have nothing against it. Have you tried it in your repo? Does it work the way you expect it to?
> 
> I'll leave it to @odsod or @viktorvoltaire to approve though

I tried it in this repository. Have not tried it in any other repository.

Makefile
```diff
@@ -8,6 +8,10 @@ sagefile := .sage/bin/sagefile
 $(sagefile): .sage/go.mod $(shell find .sage/.. -type f -name '*.go')
 	@cd .sage && go mod tidy && go run .
 
+.PHONY: build-sage
```
odsod, line 11: I have a few thoughts here:

1) Building the sagefile binary is so quick now that we have no dependencies that I think we should make it a PHONY target and always rebuild it
2) I think that this target should simply be an alias for the $(sagefile) target, since they do the same thing

Holmus: @viktorvoltaire could you chime in here?

viktorvoltaire: I like how you are thinking @odsod just tested it now and building takes half a second :) lets go for it! @Holmus sure I can push a fix

viktorvoltaire: Fixed! @Holmus @odsod @Edholm  what do think of the name `build-sage` ? another option could just be `make sage` :)

Holmus: I kinda like `make sage`

Edholm: :heavy_plus_sign: on `make sage`

viktorvoltaire: sage it is!

viktorvoltaire:
Even though I can approve this, I think its better if @Edholm and @odsod does it :rofl:

Edholm (approved):
LGTM!

odsod (approved):
Sweet

odsod:
@viktorvoltaire Should we add an `update-sage` target while we're at it...? Breaking changes and convenience considered

viktorvoltaire:
> Breaking changes and convenience considered

What do you purpose this target does?

viktorvoltaire:
I will merge this, and we can look into the update command later

PR #157, opened 2022-02-02
feat(api-linter): add Run helper

odsod:
Runs the API linter on all Buf modules beneath the current working
directory.


Edholm (approved):
LGTM!

PR #160, opened 2022-02-03
exclude results from sage folder

viktorvoltaire:
Excludes certain results from the running golangcilint in .sage. Whats good is that this does not interfere with the anyone passing their own `--exclude` as args

### DIFF
--- a/tools/sggolangcilint/tools.go
+++ b/tools/sggolangcilint/tools.go
@@ -53,7 +53,12 @@ func Run(ctx context.Context, args ...string) error {
 		if err != nil {
 			return err
 		}
-		cmd := Command(ctx, append([]string{"run", "-c", configPath, "--path-prefix", pathPrefix}, args...)...)
+		var excludeArg []string
+		if filepath.Dir(path) == sg.FromSageDir() {
+			excludeArg = append(excludeArg, "--exclude", "(is a global variable|is unused)")
+		}
+		cmdArgs := append([]string{"run", "-c", configPath, "--path-prefix", pathPrefix}, args...)
+		cmd := Command(ctx, append(cmdArgs, excludeArg...)...)
 		cmd.Dir = filepath.Dir(path)
 		commands = append(commands, cmd)
 		return cmd.Start()


Edholm (approved):
Neat! :star: 

I tried to get something similar to work locally in my sage install, but didn't work...

Have you tested it? :)

viktorvoltaire:
> Neat! star
> 
> I tried to get something similar to work locally in my sage install, but didn't work...
> 
> Have you tested it? :)

yes, tested it with :)
```golang

var iamglobal = "hello"

func BB(ctx context.Context) error {
	println(iamglobal)
	return nil
}
```

tools/sggolangcilint/tools.go
```diff
@@ -53,6 +53,9 @@ func Run(ctx context.Context, args ...string) error {
 		if err != nil {
 			return err
 		}
+		if filepath.Dir(path) == sg.FromSageDir() {
+			args = append(args, "--exclude", "(is a global variable|is unused)")
```
alethenorio, line 57: Is this omitting the output from the terminal? Shouldn't we then turn off those linters instead?

viktorvoltaire: Unfortunately we cant combine `--disable-all` and `--disable`.

alethenorio: But we control the linter config. We can just remove this from the config right?

https://github.com/einride/sage/blob/200e20f34b6e8a0d1f3819e0647b1f5d064a39b9/tools/sggolangcilint/golangci.yml#L181

alethenorio: And the same for the unused linter

alethenorio: But I also think those are good linters so I am not sure I understand the reason behind the PR (Or did we discuss reasons for removing them elsewhere?)

viktorvoltaire: Thats the thing, we dont want to remove them from the default config. Only for the `.sage` folder, given that vars is common there, and so is unused methods since thats the default for any Sage target

tools/sggolangcilint/tools.go
```diff
@@ -53,7 +53,12 @@ func Run(ctx context.Context, args ...string) error {
 		if err != nil {
 			return err
 		}
-		cmd := Command(ctx, append([]string{"run", "-c", configPath, "--path-prefix", pathPrefix}, args...)...)
+		var excludeArg []string
```
viktorvoltaire, line 56: @Edholm made a blunder in my first commit, appending to the `args` slice made it so it was applied for all runs of the command, not just the one in the sage folder :) Please have a look again

tools/sggolangcilint/tools.go
Edholm: I tried excluding deadcode linter in a specific folder, but it didn't take. I think it is because of how golangci-lint is invoked. So I think this is a good solution to the problem

Edholm (approved):
Nice!

tools/sggolangcilint/tools.go
viktorvoltaire: Yep, what @Edholm said. We invoke `golangcilint` from the root of the `go.mod`. So excluding by folder work unfortunately

alethenorio: Ah I see it now. Bummer.

alethenorio (approved):

PR #161, opened 2022-02-03
feat: rename sg.Function to sg.Target

odsod:
To add more specificity to the vocabulary (borrowing from Make
terminology) - it's hard to be specific when talking about "functions" -
it's easier to be specific when saying "Sage Targets" - because then you
know you are talking about a function that is meant to be run as a
target.


alethenorio (approved):
Great idea

PR #162, opened 2022-02-04
chore(sggo): add explanation for the TestCommand

stenehall:
This adds an explanation for the command and breaks out the
PrepareCommand parts of the command.

### DIFF
--- a/tools/sggo/tools.go
+++ b/tools/sggo/tools.go
@@ -9,6 +9,7 @@ import (
 	"go.einride.tech/sage/sg"
 )
 
+// TestCommand runs the default go test command and generates a coverage report.
 func TestCommand(ctx context.Context) *exec.Cmd {
 	coverFile := sg.FromBuildDir("go", "coverage", "go-test.txt")
 	if err := os.MkdirAll(filepath.Dir(coverFile), 0o755); err != nil {


stenehall:
Not sure if I'm doing this right :) I couldn't really see that there was any tests for this? It all seems to build with this change but if it's working is another story. The documentation does work (tested locally).

tools/sggo/tools.go
```diff
@@ -28,3 +27,9 @@ func TestCommand(ctx context.Context) *exec.Cmd {
 		"./...",
 	)
 }
+
+func PrepareCommand(_ context.Context, coverFile string) error {
```
Edholm, line 31: I would prefer to inline this, because while this package doesn't have any other functions right now it may in the future. Then this functions would seem to be relevant for all functions in the package.

My 2 cents.

viktorvoltaire: +1, we usually only have a public prepare function when we download tools. It provides an option to just download the tool as a dependency, so it can be used by other tools or outside Sage

stenehall: I don't really have any strong opinions I just tried to do it as it seemed to be done in other places, but maybe there's a difference I'm missing? (I'm new to Go)

viktorvoltaire: We lack some documentation on some of the conventions we use in Sage.  It’s on my todo! Hopefully it will make stuff more clear ☺️

stenehall: And now Viktors comment showed up after I've written mine. So download is done in prepare commands, other stuff inlined in the Command. Makes sense.

viktorvoltaire: > I don't really have any strong opinions I just tried to do it as it seemed to be done in other places, but maybe there's a difference I'm missing? (I'm new to Go)

The difference is that this tool, does not need a prepare function because it’s not downloading any external binary. That’s the main use case for prepare. Download, unpack and symlink to the bin directory.

viktorvoltaire (approved):

Edholm (approved):
Gogogo!

PR #171, opened 2022-02-08
feat(hadolint): simplify Run helper

odsod:


viktorvoltaire (commented):
This removes some flexibility from the helper, is that something we want across the board? I think we should strive for a similar implementation for other helpers in that case. So its easy to understand the logic when adding new stuff

odsod:
Yes, the golangci-lint, api-linter, clang-format and other helpers work the same way already

viktorvoltaire (approved):

PR #172, opened 2022-02-08
fix: use golang-migrate binary instead of go built version

alethenorio:
Building golang migrate requires some extra flags in order to include the spanner drivers so the current one is broken and besides that there is no reason to build it from Go since we don't have any dependencies on the library at runtime and it is generally faster to use the binary


odsod (approved):
Nice one!!

Edholm (approved):
Nice find!

PR #174, opened 2022-02-09
Add GitHub action for setting up the Go toolchain

viktorvoltaire:
With this, we can use something like

```
    - name: Set up Go Tooling
      uses: einride/sage
        cacheKey: build
```

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -12,10 +12,10 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Set up Go
-        uses: actions/setup-go@v2.1.5
+      - name: Setup Go toolchain
+        uses: ./actions/setup-go
         with:
-          go-version: ^1.17
+          cacheKey: build
 
       - name: Make
         run: make
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -14,10 +14,10 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Set up Go
-        uses: actions/setup-go@v2.1.5
+      - name: Setup Go toolchain
+        uses: ./actions/setup-go
         with:
-          go-version: ^1.17
+          cacheKey: release
 
       - name: Make
         run: make
--- a/actions/setup-go/action.yml
+++ b/actions/setup-go/action.yml
@@ -1,0 +1,37 @@
+name: Setup Go Tooling
+description: Setup Go tooling including cached Go modules.
+
+inputs:
+  cacheKey:
+    description: Custom cache key used
+    required: false
+    default: ci
+
+runs:
+  using: composite
+
+  steps:
+    - name: Set up Go
+      uses: actions/setup-go@v2.1.5
+      with:
+        go-version: ^1.17
+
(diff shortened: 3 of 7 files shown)


Edholm (approved):
woopwoop!

Edholm:
This allows for much simpler (and faster) GitHub action workflows :man_dancing:

odsod:
Do we need to work with nampespacing a little bit here?

odsod:
Like sage/actions/setup-go?

Edholm:
> - name: Set up Go Tooling
>       uses: einride/sage
>         cacheKey: build

Do you mean this @odsod ?

It isn't immediately apparent which action is being invoked by this. Dunno if we even can improve it? :thinking:

Edholm:
Musn't the action.yaml file be put under .github/actions/<name>/action.yml ? :thinking:

viktorvoltaire:
> Like sage/actions/setup-go?

Sure, I was a bit uncertain first what was allowed. But seems like any path would work. Agreed that `uses: einride/sage/actions/setup-go` is much better

viktorvoltaire:
> Musn't the action.yaml file be put under .github/actions//action.yml ? thinking

the `uses' attribute must be a path, a Docker image, or owner/repo@ref

action.yml
```diff
@@ -0,0 +1,37 @@
...
+        restore-keys: |
+          ${{ runner.os }}-go-${{ inputs.cacheKey }}-
+
+    - name: Cache sage folders
+      uses: actions/cache@v2
+      with:
+        path: |
+          ./.sage/tools
+          ./.sage/bin
+        key: ${{ runner.os }}-sage-${{ inputs.cacheKey }}-${{ hashFiles('./.mage/go.sum') }}
```
stenehall, line 35: should this be `./.sage/`?

viktorvoltaire:
I can try and make a prerelease of this branch, and try different paths @odsod @Edholm and we can see if it works

viktorvoltaire:
@odsod @Edholm this worked! Should we go with that?

```
      - name: Setup Go toolchain
        uses: einride/sage/actions/setup-go@v0.82.0-prerelease
        with:
          cacheKey: build
```

odsod:
@viktorvoltaire Nice - do we need to manually specify a cache key by default?

viktorvoltaire:
> @viktorvoltaire Nice - do we need to manually specify a cache key by default?

No its optional
```
    description: Custom cache key used
    required: false
    default: ci
```

viktorvoltaire:
For this repo, i guess we could skip it given that we just run make

action.yml
Edholm: We only want to cache generated files. We don't want to cache the go files themselves :)

Edholm:
> @viktorvoltaire Nice - do we need to manually specify a cache key by default?

The reason I added it originally is that if you have multiple jobs having the same action, they end up having the same cache key which can conflict. 

I.e. you can end up in situation where the cache for a job is incomplete because it was created by a different job.

Edholm (approved):
woop woop :+1:

Edholm (approved):

PR #176, opened 2022-02-09
feat(actions): rename to setup and include git checkout

odsod:


.github/workflows/go.yml
```diff
@@ -8,14 +8,11 @@ jobs:
...
-      - uses: actions/checkout@v2.4.0
-        with:
-          fetch-depth: 0
+      - name: Checkout
+        uses: actions/checkout@v2.4.0
 
-      - name: Setup Go toolchain
-        uses: ./actions/setup-go
-        with:
-          cacheKey: build
```
Edholm, line 18: Build and release aren't using the same go modules and thus have different requirements for the cache. I would recommend keeping these.

Imagine if `release` ends up creating the cache, then `build` will use the cache created by `release` and will probably have to download either sage tools and/or go modules every time. The way github cache works is that if there is a cache hit, it won't update the cache even if things have changed.

actions/setup/action.yml
```diff
@@ -1,20 +1,35 @@
...
 runs:
   using: composite
 
   steps:
+    - name: Checkout
+      uses: actions/checkout@v2.4.0
+      with:
+        # Needed for conventional commit linting.
+        fetch-depth: ${{ inputs.fetch-depth }}
+
```
Edholm, line 28: Having the checkout as the first step works because the action is defined in "another repo". Actions in the same repo cannot have the checkout as the first step, as you probably saw above :blush:

Edholm (approved):

Edholm:
Now @viktorvoltaire will rue the day he referenced master in the AD repo :joy:

actions/setup/action.yml
odsod: Yup! I added a little workaround

.github/workflows/go.yml
Edholm: In this repo this is most likly not an issue or if it is, a small one.

In other repos with multiple jobs I definitively recommend having unique cache keys per job.

odsod: Ah, darn - good point

actions/setup/action.yml
odsod: What if we used the workflow name as the default cache key?

Edholm: the problem is that it needs to be per job and not workflow. But the default key could obviously be the workflow name.

Then for cases where you only have a single job in a workflow, you can omit it :)

odsod: Good point again - what about `    default: ${{ github.workflow }}.${{ github.job }}`

Edholm: if that works, 100% yes

Edholm: image

It worked beautifully

Edholm (approved):
Nice!

viktorvoltaire:
Veeeery nice!

PR #180, opened 2022-02-10
feat: rename golangci-lint target to go-lint

odsod:
The specific linter may not be the same, but we'll still want to run a
standard Go linter.


Edholm (approved):
Small breaking change, but I'm fine with that :)

odsod:
It actually only impacts newly created repos!

Edholm:
ah yes ofc!

PR #181, opened 2022-02-10
feat(cloud-endpoints): add helper functions

viktorvoltaire:


tools/sgapigateway/tools.go
```diff
@@ -0,0 +1,99 @@
...
+		"%s-docker.pkg.dev/%s/docker/endpoints-runtime-serverless:%s-%s",
+		gcpRegion,
+		gcpProject,
+		serviceName,
+		configID,
+	)
+	if err := os.WriteFile(sg.FromToolsDir("api-gateway/Dockerfile"), dockerfile, 0o600); err != nil {
+		panic(err)
+	}
+	cmd := sg.Command(ctx, "docker", "build", "-t", tag, ".")
```
alethenorio, line 93: Would we benefit from using a standalone tool here instead of docker which we can download such as [buildah](https://buildah.io/) or Kaniko?

tools/sgapigateway/tools.go
```diff
@@ -0,0 +1,99 @@
...
+		gcpProject,
+		"--limit",
+		"1",
+		"--format",
+		"value(id)",
+	))
+}
+
+func DockerImage(ctx context.Context, serviceName, gcpProject, gcpRegion string) string {
+	if err := os.MkdirAll(sg.FromToolsDir("api-gateway"), 0o755); err != nil {
```
alethenorio, line 41: Can we avoid doing this in a folder which may look like it belongs in the project just for building a temporary image? Maybe do it in a temporary disposable folder such as `os.TempDir`?

alethenorio (approved):
Some small nitpicks.

tools/sgapigateway/tools.go
```diff
@@ -0,0 +1,99 @@
...
+	"net/http"
+	"os"
+
+	"go.einride.tech/sage/sg"
+)
+
+//go:embed Dockerfile
+var dockerfile []byte
+
+func EndpointsConfigID(ctx context.Context, serviceName, gcpProject string) string {
```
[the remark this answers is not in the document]
viktorvoltaire: Yes

tools/sgapigateway/tools.go
viktorvoltaire: Sure, i dont see why not!

tools/sgapigateway/tools.go
viktorvoltaire: Coolio, lets but it on the backlog to add a tool going forward :)

tools/sgapigateway/tools.go
viktorvoltaire: Went with that and created a dir i `os.TempDir`

alethenorio (approved):

PR #182, opened 2022-02-13
feat(file): add http header option to remote file downloads

andersastrom:
It would be great to be able to use this tool for file downloads that for instance require authentication through headers

### DIFF
--- a/sgtool/file.go
+++ b/sgtool/file.go
@@ -40,11 +40,13 @@ type fileState struct {
 	archiveFiles map[string]string
 	skipFile     string
 	symlink      string
+	httpHeader   http.Header
 }
 
 func newFileState() *fileState {
 	return &fileState{
 		archiveFiles: make(map[string]string),
+		httpHeader:   make(http.Header),
 	}
 }
 
@@ -65,7 +67,7 @@ func FromRemote(ctx context.Context, addr string, opts ...Opt) error {
 		}
 	}
 	sg.Logger(ctx).Printf("fetching %s...", addr)
-	rStream, cleanup, err := downloadBinary(ctx, addr)
+	rStream, cleanup, err := s.downloadBinary(ctx, addr)
 	if err != nil {
 		return fmt.Errorf("unable to download file: %w", err)
 	}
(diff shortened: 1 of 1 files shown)


odsod (changes requested):
This looks like it would work better as an option instead of adding a required parameter and passing `nil` most of the time.

andersastrom:
> This looks like it would work better as an option instead of adding a required parameter and passing `nil` most of the time.

Ah yeah that makes sense! 👍🏼

sgtool/file.go
```diff
@@ -48,7 +48,7 @@ func newFileState() *fileState {
 	}
 }
 
-func FromRemote(ctx context.Context, addr string, opts ...Opt) error {
+func FromRemote(ctx context.Context, addr string, header http.Header, opts ...Opt) error {
```
alethenorio, line 51: How about introducing some functional options for it?

alethenorio:
Oops just saw @odsod comment 😅

odsod:
@andersastrom Might as well take multiple headers into account then also? Make the option a varargs where multiple headers can be supplied?

odsod:
@andersastrom Just realized that http.Header is actually an alias for multiple headers - disregard me!

andersastrom:
> @andersastrom Just realized that http.Header is actually an alias for multiple headers - disregard me!

yes that confused me as well 😅

viktorvoltaire (approved):
Nice! LGTM

sgtool/file.go
```diff
@@ -189,11 +191,19 @@ func WithSkipIfFileExists(filepath string) Opt {
 	}
 }
 
-func downloadBinary(ctx context.Context, url string) (io.ReadCloser, func(), error) {
+func WithHTTPHeader(header http.Header) Opt {
```
Edholm, line 194: Would it make sense to have this as a key value and then append to the already allocated f.httpHeader? Then you can pass as many opts as you want :)

This option will overwrite any previous headers :)

andersastrom: Just so I understand you correctly, do you mean passing each key value pair individually rather than the full map?
i.e. something like
```golang
.WithHTTPHeader("k", "v")
.WithHTTPHeader("k", "v")
...
```

Edholm: Yep exactly :+1: 

You don't need to change, just consider it and see how you feel about it :)

andersastrom: Cool! I like it, I don't have a strong preference but I think its neater if you only need to pass small number of headers (which probably is the case most of the time) :)

Edholm: Yeah, if you want you can have both options I guess :shrug: 
`WithHTTPHeaders(<map>)` and `WithHTTPHeader(key,val)` 

But I think the single header use-case is probably going to be the the most used one

Edholm (approved):
Neat :100:

sgtool/file.go
```diff
@@ -189,11 +191,19 @@ func WithSkipIfFileExists(filepath string) Opt {
 	}
 }
 
-func downloadBinary(ctx context.Context, url string) (io.ReadCloser, func(), error) {
+func WithHTTPHeader(key, value string) Opt {
```
odsod, line 194: This API seems a bit limiting (what about multiple values for the same header?) - why not do `http.Header` as argument?

andersastrom: As [mentioned](https://github.com/einride/sage/pull/182#discussion_r805593612) I don't have a strong preference. Adding multiple values for the same header will still work as long as we do `f.httpHeader.Add(key, value)` rather than use `.Set()`

Edholm: hehe that is what he used originally :D

andersastrom: I agree that for this specific scenario however that is not very obvious simply looking at the API though

odsod (approved):
Coolio

PR #187, opened 2022-02-18
feat(sggit): expose git as a command

thall:
In the driver-app we generate tags with `git describe --tags --long --always`, so would be useful to expose git as a command to follow the same pattern as the other.

And we are using `sggit.VerifyNoDiff` and would like to not have mix of `sggit` and `sg.Command(ctx, "git", ...)`


alethenorio (approved):
Nice.

PR #189, opened 2022-02-21
feat: remove go dependency for every build step

sharkema:
Purpose:
Remove a go dependency for every build step in cloud build. 

Description:
With the 'old' setup, sage will require a dependency on 'go' for every buildstep in cloud-build. This gives more complexities than needed. Every non-go buildstep will need a image with both the desired language as well as go, as sage depends on this. For example if you want to build something with the node could-builder image, it will fail as it has a dependency on go. 

Solution:
The updated state will allow for non-go applications to first call `make sage` with a go based cloud-builder image, which will generate the binary sage file. And subsequently build the app with the desired default type of building images.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,7 @@ $(sagefile):
 	@cd .sage && go mod tidy && go run .
 
 .PHONY: sage
+sage: $(shell git clean -fxq $(sagefile))
 sage: $(sagefile)
 
 .PHONY: update-sage
--- a/sg/generate.go
+++ b/sg/generate.go
@@ -14,6 +14,7 @@ import (
 // GenerateMakefiles defines which Makefiles should be generated.
 func GenerateMakefiles(mks ...Makefile) {
 	ctx := WithLogger(context.Background(), NewLogger("sage"))
+	Logger(ctx).Println("building binary and generating Makefiles...")
 	if len(mks) == 0 {
 		panic("no makefiles to generate, see https://github.com/einride/sage#readme for more info")
 	}
@@ -67,7 +68,7 @@ func GenerateMakefiles(mks ...Makefile) {
 		mk := codegen.NewMakefile(codegen.FileConfig{
 			GeneratedBy: "go.einride.tech/sage",
 		})
-		if err := generateMakefile(mk, pkg, v, mks...); err != nil {
+		if err := generateMakefile(ctx, mk, pkg, v, mks...); err != nil {
 			panic(err)
 		}
 		if err := os.WriteFile(v.Path, mk.RawContent(), 0o600); err != nil {
--- a/sg/makefile.go
+++ b/sg/makefile.go
@@ -1,6 +1,7 @@
 package sg
 
 import (
+	"context"
 	"fmt"
 	"go/ast"
 	"go/doc"
@@ -43,7 +44,7 @@ func (m Makefile) defaultTargetName() string {
 	return result
 }
 
-func generateMakefile(g *codegen.File, pkg *doc.Package, mk Makefile, mks ...Makefile) error {
+func generateMakefile(ctx context.Context, g *codegen.File, pkg *doc.Package, mk Makefile, mks ...Makefile) error {
 	includePath, err := filepath.Rel(filepath.Dir(mk.Path), FromSageDir())
 	if err != nil {
 		return err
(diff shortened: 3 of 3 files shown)


Makefile
```diff
@@ -5,11 +5,11 @@
 
 sagefile := .sage/bin/sagefile
 
-.PHONY: $(sagefile)
```
viktorvoltaire, line 8: I understand why you want to make this change. Some background: We made this a phony target because regenerating the binary was so fast. If we are to remove the phony, we need to reintroduce the file dependencies that this target had, which was `go.sum` and `*.go` files in the `.sage` folder.

Otherwise the binary will never be updated when running locally.

viktorvoltaire: You can find the exact stuff here https://github.com/einride/sage/commit/b9faf866ef21fc2b05f81b8e8220300f369b8bd1 :)

Makefile
```diff
@@ -5,11 +5,11 @@
...
-$(sagefile):
+$(sagefile):.sage/go.mod $(shell find .sage/.. -type f -name '*.go')
 	@cd .sage && go mod tidy && go run .
 
 .PHONY: sage
+sage: clean-sage
```
odsod, line 12: Do we really want to re-download all the tools everytime we rebuild the sagefile binary?

Makefile
odsod: missing a space after `$(sagefile):` here

Makefile
viktorvoltaire: No I dont think we should

Makefile
viktorvoltaire: @odsod @sharkema should we also bring back the logging, which we had before, when regenerating the binary. It was unnecessary when we regenerated each time. But now I think its nice to see when it happens

Makefile
sharkema: :+1:  Updated such that it is cleaning only the sagefile binary, such that `make sage` forces a recreate.

Makefile
sharkema: Currently it logs on generation of the binary:
```
```

viktorvoltaire: Oh, I thought we removed that. No then we are good

odsod: I thought it was a little bit noisy

Makefile
```diff
@@ -5,17 +5,21 @@
...
 .PHONY: sage
+sage: clean-sage-file
 sage: $(sagefile)
 
 .PHONY: update-sage
 update-sage:
 	@cd .sage && go get -d go.einride.tech/sage@latest && go mod tidy && go run .
 
+.PHONY: clean-sage-file
+clean-sage-file:
```
viktorvoltaire, line 20: Feels like this can just be inlined in the `sage` target?

Makefile
viktorvoltaire: Yes, but given that we wont recreate it by default, its might be more clear to the user if its now printed?

odsod: Sounds good to me!

Makefile
odsod: aye

Makefile
sharkema: I see my confusion, it was removed here indeed; https://github.com/einride/sage/commit/991a4cc2f4b6d11f768862e063f379ebe4644129  though the branch I was working on was depending on an older version

Makefile
sharkema: :+1: I used one way I found in the docs of make to realize it, but I am not sure if this is the preferred way

viktorvoltaire (approved):
Sweet! LGTM 👌

Makefile
```diff
@@ -5,11 +5,11 @@
 
 sagefile := .sage/bin/sagefile
 
-.PHONY: $(sagefile)
-$(sagefile):
+$(sagefile): .sage/go.mod $(shell find .sage/.. -type f -name '*.go')
```
odsod, line 8: Why the `..` after `.sage` - is that intentional?

sharkema: I added what was formerly there. (originally added in https://github.com/einride/sage/commit/296ceab0d84a66543506b5374d11dc1bc92bc47c#diff-23e713a554f536f060bdf56c6960e598d68657017b92b1466b49339ec040f061R2 ) . My feel is that as this is the sage repository, there should be more go files one level up. 

I honestly find it hard to argue why `$(shell find .sage/.. -type f -name '*.go')` will add any value. If find does not resolve any files, it will be expanded to nothing, which is fine for make. If find does resolve anything those files will by definition be there. 

What is it I am missing? or shall we remove that clause?

sharkema: @odsod What do you think?

odsod: Hmm, I was thinking that it's only the Go files in the `.sage` module that impact if the need to rebuild the sagefile. The actual source code changing should not trigger a rebuild of the build tooling - or am I missing something? So it should `find` from the root of the sage module, but not one level above?

viktorvoltaire: We look for all go files in the repository for this repository only, it's a conditional in the generate code. This is because this repo is special :) and as you can see in the go mod in the sage folder, we have a local replace.

We could simplify it by just always rebuilding in this repo. So rewrite the conditional in the generate file to add a Phony target instead.

sharkema: :+1: makes sense, cleared the dots.

viktorvoltaire: Great!

PR #190, opened 2022-02-21
fix(sg): support filepaths from both git root and current working directory

thall:


viktorvoltaire (approved):
Perfect!

odsod (approved):
Good in theory, but I don't think GitHub will automatically detect file paths from the current working directory. Which tool is it that outputs file paths like this?

thall:
> Good in theory, but I don't think GitHub will automatically detect file paths from the current working directory. Which tool is it that outputs file paths like this?

We had this problem in driver-app, https://github.com/einride/driver-app/runs/5271396785?check_suite_focus=true, golangcilint reports full path from git-root but our code checks from current working directory

viktorvoltaire:
> Good in theory, but I don't think GitHub will automatically detect file paths from the current working directory. Which tool is it that outputs file paths like this?

the `os.Lstat` gets run from the current working directory in this case, so when running `make -C someDir go-lint` the command command will fail, because the `pathPrefix` is calculated from `filepath.Rel(sg.FromGitRoot(), filepath.Dir(path))` where `path` is the path to the `go.mod`

sg/exec.go
```diff
@@ -40,7 +41,7 @@ func (l *logWriter) Write(p []byte) (n int, err error) {
 	for in.Scan() {
 		line := in.Text()
 		if !l.hasFileReferences {
-			l.hasFileReferences = hasFileReferences(line)
+			l.hasFileReferences, line = hasFileReferences(line)
```
viktorvoltaire, line 44: I think this will mess with indentation now, since we do trimSpace in `hasFileReferences`

thall:
Solution now is to always resolve file path to be relative from root, to account for programs which only outputs paths relative from CWD.

viktorvoltaire (approved):
LGTM

sg/exec.go
```diff
@@ -62,14 +61,23 @@ func (l *logWriter) Write(p []byte) (n int, err error) {
 	return len(p), nil
 }
 
-func hasFileReferences(line string) bool {
-	line = strings.TrimSpace(line)
-	if i := strings.IndexByte(line, ':'); i > 0 {
-		if _, err := os.Lstat(line[:i]); err == nil {
-			return true
+func hasFileReferences(line string) (bool, string) {
```
odsod, line 64: this function needs a new name now that it does more stuff?

sg/exec.go
```diff
@@ -62,14 +61,30 @@ func (l *logWriter) Write(p []byte) (n int, err error) {
 	return len(p), nil
 }
 
-func hasFileReferences(line string) bool {
-	line = strings.TrimSpace(line)
-	if i := strings.IndexByte(line, ':'); i > 0 {
-		if _, err := os.Lstat(line[:i]); err == nil {
-			return true
+func hasFileReferenceAndEnsurePathRelativeToGitRoot(line string) (bool, string) {
```
thall, line 64: not happy with this one, any suggestions?

PR #194, opened 2022-02-23
Fix/make sage

sharkema:
The git clean command needs to be part of the recipe. Otherwise variable expansion will evaluate the command always, independing of the target called. :bulb: :see_no_evil:


viktorvoltaire (approved):
😁

odsod (approved):

PR #195, opened 2022-02-24
feat(exec): add ContextWithEnv func

viktorvoltaire:


sg/exec.go
```diff
@@ -13,13 +13,31 @@ import (
...
 	cmd.Dir = FromGitRoot(".")
-	cmd.Env = prependPath(os.Environ(), FromBinDir())
+	if v := ctx.Value(cmdEnvKey); v != nil {
+		env, ok := v.([]string)
+		if ok {
+			cmd.Env = append(os.Environ(), env...)
+		}
+	} else {
+		cmd.Env = os.Environ()
+	}
```
alethenorio, line 39: Nitpick

```suggestion
 cmd.Env = os.Environ()
	if v := ctx.Value(cmdEnvKey); v != nil {
		env, ok := v.([]string)
		if ok {
			cmd.Env = append(cmd.Env, env...)
		}
	}
```

alethenorio (approved):
🎉🎉

sg/exec.go
viktorvoltaire: Ah yes, better!

PR #196, opened 2022-02-24
feat(sggcloud): add GHA WIF auth helper

viktorvoltaire:
This introduces a replacement for handling auth in the actions pipeline, rather then having the below, one can now just use this function and pass the returned json file in the context with `sg.ContextWithEnv` to what ever command that needs it.
```yaml
    - id: 'auth'
        name: 'GCP Auth'
        uses: 'google-github-actions/auth@v0'
        with:
          workload_identity_provider: 'projects/PROJECT_NR/locations/global/workloadIdentityPools/github-actions/providers/github-actions'
          service_account: 'someServiceAccount@...'
```

### DIFF
--- a/tools/sggcloud/tools.go
+++ b/tools/sggcloud/tools.go
@@ -1,0 +1,32 @@
+package sggcloud
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"go.einride.tech/sage/sg"
+)
+
+// GHACredentials generates short-lived credentials for GCP with workload identity federation.
+func GHACredentials(ctx context.Context, wifPoolName, serviceAccount string) string {
+	outputFile := filepath.Join(os.TempDir(), fmt.Sprintf("gha-creds-%s.json", strings.Split(wifPoolName, "/")[1]))
+	if _, err := os.Stat(outputFile); err == nil {
+		return outputFile
+	}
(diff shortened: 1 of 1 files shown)


tools/sggcloud/tools.go
```diff
@@ -0,0 +1,33 @@
...
+	if _, err := os.Stat(outputFile); err == nil {
+		return outputFile
+	}
+	//nolint:lll
+	wifPool := fmt.Sprintf("projects/%s/locations/global/workloadIdentityPools/github-actions/providers/github-actions", projectNumber)
+	audienceURL := fmt.Sprintf("https://iam.googleapis.com/%s", wifPool)
+	if err := sg.Command(ctx, "gcloud", "iam", "workload-identity-pools", "create-cred-config",
+		wifPool,
+		"--service-account", serviceAccount,
+		"--output-file", outputFile,
```
alethenorio, line 24: Any chance it is possible to supply a dash here (`-`) to get the output in the stdout rather than a file so we avoid having to go through disk writes in the host for this?

alethenorio (approved):

tools/sggcloud/tools.go
viktorvoltaire: This file is read by the GOOGLE_APPLICATION_CREDENTIALS, which only accept a file

PR #198, opened 2022-02-25
feat: add support for namespace embedding

alethenorio:
I was trying to design a reusable namespace with sage and thought this might be something interesting to support.

The idea is to be able to do something like this

```go
package main

import "fmt"

type MyNamespace struct {
	sg.Namespace
	Name string
}

func (m MyNamespace) All(ctx context.Context) error {
	sg.SerialDeps(
		ctx,
		m.PrintName,
	)
	return nil
}

func (n MyNamespace) PrintName(ctx context.Context) error {
	fmt.Println(n.Name)
}
```

And then we can reuse the namespace to generate different makefiles

```go
func main() {
	sg.GenerateMakefiles(
		sg.Makefile{
			Path:          sg.FromGitRoot("Makefile"),
			DefaultTarget: All,
		},
		name1 := MyNamespace{Name: "name1"}
		sg.Makefile{
			Path:          sg.FromGitRoot("names/name1/Makefile"),
			DefaultTarget: name1.All,
			Namespace:     name1,
		},
		name2 := MyNamespace{Name: "name2"}
		sg.Makefile{
                        Path:          sg.FromGitRoot("names/name2/Makefile"),
                        DefaultTarget: name2.All,
                        Namespace:     name2,
                },
	)
}
```

The use case here was to design a repo which will contain folders with different applications each with their own makefile.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,38 @@ func (Terraform) TerraformInitDev() {
 }
 ```
 
+It is also possible to embedded a Namespace in order to add metadata to it and potentially reuse it for different Makefiles
+
+```golang
+
+func main() {
+	sg.GenerateMakefiles(
+		sg.Makefile{
+			Path:          sg.FromGitRoot("Makefile"),
+			DefaultTarget: All,
+		},
+		sg.Makefile{
+			Path:          sg.FromGitRoot("names/name1/Makefile"),
+			Namespace:     MyNamespace{Name: "name1"},
+		},
+		sg.Makefile{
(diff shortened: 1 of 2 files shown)


odsod:
It looks too good to be true! :joy: We should document this in the README though!

alethenorio:
> It looks too good to be true! joy We should document this in the README though!

Good point. Fixed

viktorvoltaire (approved):
Awesome addition!

PR #199, opened 2022-02-28
Mac M1 support for protoc-gen-go-grpc 

backevik:
- fix: mac m1 support for protocgengogrpc
- refactor: "darwin" string to constant

### DIFF
--- a/sgtool/file.go
+++ b/sgtool/file.go
@@ -29,6 +29,10 @@ const (
 )
 
 const (
+	Darwin = "darwin"
+)
+
+const (
 	AMD64 = "amd64"
 	X8664 = "x86_64"
 )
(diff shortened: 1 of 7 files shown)


sgtool/file.go
```diff
@@ -28,6 +28,10 @@ const (
 	TarGz
 )
 
+const (
+	DARWIN = "darwin"
```
odsod, line 32: This should say `Darwin`?

odsod: Since Go constants are not all upper case

backevik: Thank you! Switching between TS and Go can be confusing 😅  Fixed!

odsod (approved):

PR #200, opened 2022-02-28
fix(embedded-ns): reflect on values in embedded namesapce

viktorvoltaire:
This PR partially fixes what was introduced in https://github.com/einride/sage/pull/198

I decided only to reflect String, Int and Bool. As any other types gets very hard and complex to template, I think these are fine to start with anyway and we can see where this feature takes us

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ func (Terraform) TerraformInitDev() {
 }
 ```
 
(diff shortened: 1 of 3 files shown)


alethenorio (approved):
Awesome. Thank you

PR #201, opened 2022-03-01
fix: only include make targets in binary

viktorvoltaire:


sg/initfile.go
```diff
@@ -33,74 +37,78 @@ func generateInitFile(g *codegen.File, pkg *doc.Package, mks []Makefile) error {
...
+							}
+							i++
+						}
+					}
+					ns := "{}."
+					for _, mk := range mks {
+						if mk.namespaceName() != "" && mk.namespaceName() == function.Recv {
+							ns = getNamespaceMetadata(mk)
+						}
+					}
```
alethenorio, line 88: Should this be a loop at all? We are already inside an mk so now we are looping twice. Maybe

```suggestion
		if mk.namespaceName() != "" {
			ns = getNamespaceMetadata(mk)
		}
```

sg/initfile.go
```diff
@@ -33,74 +37,78 @@ func generateInitFile(g *codegen.File, pkg *doc.Package, mks []Makefile) error {
...
+					g.P("if err != nil {")
+					g.P("logger.Fatal(err)")
+					g.P("}")
+				} else {
+					ns := "{}."
+					for _, mk := range mks {
+						if mk.namespaceName() != "" && mk.namespaceName() == function.Recv {
+							ns = getNamespaceMetadata(mk)
+						}
+					}
```
alethenorio, line 105:
```suggestion
		if mk.namespaceName() != "" {
			ns = getNamespaceMetadata(mk)
		}
```

alethenorio (approved):
We should probably add some generated example targets. These can get difficult to review for correctness :D

sg/initfile.go
viktorvoltaire: I refactored this before I saw your comment :)

sg/initfile.go
viktorvoltaire: I refactored this before I saw your comment :)

PR #209, opened 2022-03-07
fix: run target generation function for all functions

alethenorio:
When one function was not eligible for generation, it was preventing all other functions from being evaluated in the package and namespaces.
The bool return wasn't doing much so instead we just run the function


Edholm (approved):
I can confirm this solves my issue, nice find! :100:

viktorvoltaire (approved):
Ah perfect, the bool fooled me when we updated the generation logic, I like this more 👍

PR #210, opened 2022-03-08
fix(api-linter): walk from git root instead of cwd

odsod:
Works consistently regardless of which working dir it's called from.


thall (approved):
Nice!

odsod:
Noice!!

PR #214, opened 2022-03-11
feat: remove spanner-aip-go

thall:
Since spanner-aip-go is both a compile- and runtime library it can be
dangerous to use it as an binary since you can have another version in
your repo.

In theory you could then generate code that you then can't run because of
mixed versions, and vice versa.

Use instead:
```go
return sg.Command(ctx, "go", "run", "-mod=mod",
  "go.einride.tech/spanner-aip", "generate").Run()
```


alethenorio:
@thall Spanner-aip-go can be used to generate code without using any other parts of the library in the actual application generating the code so I think this is a valid use case.

Maybe we could figure out whether the application has it in its go.mod and dynamically do the right way?

odsod (approved):
@alethenorio I see your point, but by offering the tool like this I think we're creating a bit of a pitfall for unsuspecting users.

It was great that @thall caught this now, or we might not have caught it until the code generator and runtime fell out of sync and caused bugs!

alethenorio:
Got it but how about detecting this and building the way that it needs to be built? I can give it a try and push a PR for this later

thall:
What I'm a bit hesitance about here is that it is very easy to use it wrongly, its just the mix that is confusing.

alethenorio:
Yeah I understood. I'll give it a try to detect it dynamically and building using the go.mod file if the dependency exists there and use the version otherwise

odsod:
@alethenorio What if it's a multi-module repository though? Just thinking that auto-detection might be a pitfall itself, because it might autodetect the top-level module when the version is specified in a submodule?

alethenorio:
Yeah multi-module would not be a supported use case in this case. @thall let's merge this and if I find a solution thatits good enough for this I can bring a new PR up

PR #215, opened 2022-03-11
feat: support struct args in Fn

maxekman:
Needed to support passing structs as params in the context of the github.com/einride/dn tool.


maxekman:
Will not work in make files, closing.

PR #218, opened 2022-03-15
feat: reusable namespaces

alethenorio:
This PR adds support for reusing a namespace for more than 1 Makefile essentially allowing you do to the following

```go
type Terraform sg.Namespace

func (Terraform) Plan(ctx context.Context) error {
	// do terraform plan here
}

func main() {
	sg.GenerateMakefiles(
		sg.Makefile{
			Path:          sg.FromGitRoot("Makefile"),
			DefaultTarget: All,
		},
		sg.Makefile{
			Path:          sg.FromGitRoot("project1/Makefile"),
			Namespace:     Terraform{},
		},
		sg.Makefile{
			Path:          sg.FromGitRoot("project2/Makefile"),
			Namespace:     Terraform{},
		},
	)
}
```

Which would then generate a Makefile in folder `project1` and `project2` with the targets in the Terraform namespace

The main change here is that rather than 1 single init function in the generated sage binary, we now generate 1 function for each Makefile and use the hash of the defined Makefile path as a prefix to know which Makefile is calling the sage binary.

Together with this work I refactored the generation code and broke it out into a few isolated functions to try and make that code, which was getting difficult to work with, a little easier to understand.

There is one major user facing change and a minor one.

The major one is that the targets generated in the root Makefile for calling other Makefiles is now the path of the Makefile file directory rather than the namespace name. So before, for the example above, you would have `make terraform` as a target in the root Makefile and now you have `make project1` and `make project2` instead (which I think it makes sense from a point of view of make.

The minor one is that the sage binary call inside the make target would be slight more readable and now will contain the hashed prefix. So in the example above you would have

```Makefile
.PHONY: terraform-plan
terraform-plan: $(sagefile)
	@$(sagefile) Terraform:Plan
```

and now will look something like this

```Makefile
.PHONY: terraform-plan
terraform-plan: $(sagefile)
	@$(sagefile) "Target1692162558:Terraform:Plan"
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,32 +43,32 @@ clean-sage:
 
 .PHONY: all
 all: $(sagefile)
-	@$(sagefile) All
+	@$(sagefile) "Target2901124887:All"
 
 .PHONY: convco-check
 convco-check: $(sagefile)
-	@$(sagefile) ConvcoCheck
+	@$(sagefile) "Target2901124887:ConvcoCheck"
 
 .PHONY: format-markdown
 format-markdown: $(sagefile)
-	@$(sagefile) FormatMarkdown
+	@$(sagefile) "Target2901124887:FormatMarkdown"
 
 .PHONY: git-verify-no-diff
 git-verify-no-diff: $(sagefile)
-	@$(sagefile) GitVerifyNoDiff
+	@$(sagefile) "Target2901124887:GitVerifyNoDiff"
 
 .PHONY: go-lint
 go-lint: $(sagefile)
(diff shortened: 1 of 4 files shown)


Edholm (approved):
It's a bit hard to review due to the nature of how the the makefile generation works but looks good to me at least :shrug:

Edholm:
@alethenorio should we keep the branch?

alethenorio:
> @alethenorio should we keep the branch?

I think we can always bring it back it we need to so no reason to keep it

PR #219, opened 2022-03-17
feat: add dbt tool

jonasalm:
Install dbt (data build tool) with BigQuery adapter using pip.

### DIFF
--- a/tools/sgdbt/tools.go
+++ b/tools/sgdbt/tools.go
@@ -1,0 +1,61 @@
+package sgdbt
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+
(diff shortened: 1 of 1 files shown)


Edholm (commented):
Do you see a need for this in multiple repos? The reason I ask is that since we're using pip to install the tool, it isn't really self-contained to the repo right?

jonasalm:
> Do you see a need for this in multiple repos? The reason I ask is that since we're using pip to install the tool, it isn't really self-contained to the repo right?

Yes, I see this tool being used in the Saga mesh nodes where the data pipelines can be written in SQL. There is a dependency on having `python3` in the environment though.

viktorvoltaire (approved):
Sweet

PR #220, opened 2022-03-18
feat: bump golangci linter

alethenorio:
This version includes support for Go 1.18 however it does so by disabling the following plugins

* bodyclose
* contextcheck
* gosimple
* nilerr
* noctx
* rowserrcheck
* sqlclosecheck
* staticcheck
* stylecheck
* tparallel
* unparam
* unused
* wastedassign

And requiring an extra config paramater in the config file which we do not include in this PR.

Since we are using many of these linters today, this PR is meant to allow users to bring their own local repo config with 1.18 support enabled if they so wish but keeping the 1.17 config for now

### DIFF
--- a/tools/sggolangcilint/tools.go
+++ b/tools/sggolangcilint/tools.go
@@ -17,7 +17,7 @@ import (
 
 const (
 	name    = "golangci-lint"
-	version = "1.44.2"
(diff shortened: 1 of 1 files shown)


odsod (approved):
Nice

PR #221, opened 2022-03-18
feat: enable Go 1.18 on golangci-lint per default

alethenorio:
This PR enables Go 1.18 support in GolangCI lint per default.

Draft until someone can verify it works as intended in Go 1.18

NOTE: GolangCI Lint removes the `bodyclose` comment for some reason


Edholm:
This is probably not needed since https://github.com/einride/sage/pull/227

alethenorio:
Closing it then

PR #223, opened 2022-03-23
fix(goreview): upgrade to v0.20.0 in order to support Go 1.18

Edholm:
Removes a deprecated invocation of `strings.Title` and upgrades goreview which depended on an old version of golang.org/x/tools which doesn't work properly in > Go 1.18

### DIFF
--- a/tools/sggoreview/tools.go
+++ b/tools/sggoreview/tools.go
@@ -6,13 +6,13 @@ import (
 	"os/exec"
 	"path/filepath"
 	"runtime"
-	"strings"
+	"unicode"
 
 	"go.einride.tech/sage/sg"
 	"go.einride.tech/sage/sgtool"
 )
 
-const version = "0.18.0"
+const version = "0.20.0"
 
 // nolint: gochecknoglobals
 var commandPath string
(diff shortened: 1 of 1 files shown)


tools/sggoreview/tools.go
```diff
@@ -27,7 +27,8 @@ func PrepareCommand(ctx context.Context) error {
 	toolDir := sg.FromToolsDir(toolName)
 	binDir := filepath.Join(toolDir, version, "bin")
 	binary := filepath.Join(binDir, toolName)
-	hostOS := strings.Title(runtime.GOOS)
+	goos := []rune(runtime.GOOS)
+	hostOS := string(append([]rune{unicode.ToUpper(goos[0])}, goos[1:]...)) // capitalizes the first letter.
```
Edholm, line 31: If anyone knows of a better way to do this, let me know :sweat_smile:

Edholm: The documentation states:
>Deprecated: The rule Title uses for word boundaries does not handle Unicode punctuation properly. Use golang.org/x/text/cases instead.

but I didn't want to add a dependency just for this.

odsod: we have a helper function that gets copied around quite a bit - if you search for unicode.ToUpper you'll find it :smile:

Edholm: aight, I'll leave as is for now :sweat_smile:

PR #224, opened 2022-03-23
Feat / Add sqlc tool

maxekman:
SQLC is a code generator that generate static Go structs from SQL queries. This ensures both performance and type safety for SQL, in addition to a single source of truth (operations are defined in SQL only).

https://github.com/kyleconroy/sqlc

### DIFF
--- a/tools/sgsqlc/tools.go
+++ b/tools/sgsqlc/tools.go
@@ -1,0 +1,49 @@
+package sgsqlc
+
+import (
+	"context"
+	"fmt"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+
+	"go.einride.tech/sage/sg"
+	"go.einride.tech/sage/sgtool"
+)
+
+const (
+	version = "1.12.0"
+	name    = "sqlc"
+)
+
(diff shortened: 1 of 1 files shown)


tools/sgsqlc/tools.go
```diff
@@ -0,0 +1,28 @@
...
+// nolint: gochecknoglobals
+var commandPath string
+
+func Command(ctx context.Context, args ...string) *exec.Cmd {
+	sg.Deps(ctx, PrepareCommand)
+	return sg.Command(ctx, commandPath, args...)
+}
+
+func PrepareCommand(ctx context.Context) error {
+	binary, err := sgtool.GoInstall(ctx, "github.com/kyleconroy/sqlc/cmd/sqlc", version)
```
odsod, line 22: Let's use the prebuilt binary instead? https://github.com/kyleconroy/sqlc/releases - usually faster download, especially on GitHub Actions

Edholm (approved):
Neat! Very useful tool this

tools/sgsqlc/tools.go
maxekman: Done!

PR #232, opened 2022-03-28
feat(golangci-lint): expose default config

alethenorio:
Expose the default golangci-lint config to be used by other projects who want to run the `Command()` but still want to benefit from the defaul configuration.


Edholm (approved):
Neat!

PR #234, opened 2022-03-29
fix(api-linter): add basic retry to mitigate flaky buf build

odsod:
The descriptor build step sometimes fails with context timeout - this
mitigates the flakyness.


Edholm (approved):
Let's try it!

PR #235, opened 2022-03-31
feat: rename all Makefiles on init

alethenorio:
If the repo contains several Makefile when running init, rename not only
the main root Makefile but all Makefile files in the repo

### DIFF
--- a/main.go
+++ b/main.go
@@ -7,6 +7,7 @@ import (
 	_ "embed"
 	"errors"
 	"fmt"
+	"io/fs"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


odsod:
That might not be what we want. Some repos may want a custom Makefile in a subdirectory (we have seen examples of this).

alethenorio:
> That might not be what we want. Some repos may want a custom Makefile in a subdirectory (we have seen examples of this).

Yes but this is only on init. We are not touching any custom Makefiles created after the repo has migrated to sage

alethenorio:
But this is mostly an idea. What this is meant to fix is to avoid writing a namespace to generate a different Makefile and then running sage which will nuke the existing Makefile but it could also be seen as an implementation error by the user not taking care to back the previous Makefile

ericwenn:
Personally, I don't think this is needed, and is hard to do in practice (would not want to scan the full `node_modules` folder for example).

`sage init` generates a replacement for the top-level `Makefile`, so it makes sense that it does not overwrite the existing `Makefile`. Unless we automatically infer what namespaces are needed (dont think this is a good idea), we should not backup non-top-level `Makefiles` either.

stenehall:
>

stenehall:
> Personally, I don't think this is needed, and is hard to do in practice (would not want to scan the full `node_modules` folder for example).
> 
> `sage init` generates a replacement for the top-level `Makefile`, so it makes sense that it does not overwrite the existing `Makefile`. Unless we automatically infer what namespaces are needed (dont think this is a good idea), we should not backup non-top-level `Makefiles` either.

This will likely not be a problem for us at Einride since we always have the Makefiles committed in git. However nuking existing Makefiles for someone else might cause someone some head-ache. I ran into this when moving some frontend code around and while I realise this isn't weird it's something I didn't think about when I ran the sage command while porting my Makefiles into sage.

alethenorio:
Further conversations IRL lean towards this not being the right approach to solve the issue at hand so I am closing it for now.

PR #240, opened 2022-04-05
feat(actions): pin go-version to 1.17

thall:


Edholm:
Are there issues with 1.18 or why the downgrade? :)

Edholm:
Ah, what do you think about actively setting 1.18 instead? Forwards instead of backwards? 

I've been using 1.18 and it works fine now after a few updates to golangci-lint and goreview

thall:
> Are there issues with 1.18 or why the downgrade? :)

A combination :) 

- Failing [job](https://github.com/einride/driver-app/runs/5828488928?check_suite_focus=true).
- And that we upgrade without us knowing about it. I don't see the reason why we have it like that.

thall:
> Ah, what do you think about actively setting 1.18 instead? Forwards instead of backwards?
> 
> I've been using 1.18 and it works fine now after a few updates to golangci-lint and goreview

I'm fine with both 1.17 and 1.18 as long its fixed :) Is 1.18 available for older Ubuntu releases? Just thinking of developers that may not yet be up to date.

alethenorio:
I think it makes sense to not force a new version on every developer until they have had a chance to fix any issues related to it in their code base.

I think it makes sense to pin to a specific version and then have a PR that clearly bumps the Go version instead of relying on autobump.

Maybe even include a warning message a couple of releases before about the impending change?

alethenorio:
@thall I understand your use case is due to the change to deprecate the `strings.Title` in Go 1.18. Is it a large change to make or could we potentially fix this on your side for now and maybe bump to a fixed 1.18 in this PR?

Edholm:
ah I'm also fine with both.

Go 1.18 is available in devtools.

Edholm (approved):
Let's merge and maybe set a date where we bump to 1.18?

thall:
> Let's merge and maybe set a date where we bump to 1.18?

I think that make sense. Should be a topic for next infra-guild.

thall:
> @thall I understand your use case is due to the change to deprecate the `strings.Title` in Go 1.18. Is it a large change to make or could we potentially fix this on your side for now and maybe bump to a fixed 1.18 in this PR?

Yes, that's the symptom, what i'm after is that we upgrade without knowing about it.

PR #242, opened 2022-04-07
feat: add function to handle local archive files

alethenorio:
Was originally removed because it is not used in this repo however it does not mean it is not used in other repos and is quite a useful function to have for working with local archive files for tools not yet introduced in this repo

### DIFF
--- a/sgtool/file.go
+++ b/sgtool/file.go
@@ -55,6 +55,33 @@ func newFileState() *fileState {
 	}
 }
 
+// FromLocal can be used to work with local archive files.
+// HTTP related Options, such as WithHTTPHeader don't do anything here.
+func FromLocal(ctx context.Context, filepath string, opts ...Opt) error {
+	s := newFileState()
+	for _, o := range opts {
+		o(s)
+	}
+	if s.skipFile != "" {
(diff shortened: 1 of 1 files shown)


sgtool/file.go
```diff
@@ -55,6 +55,31 @@ func newFileState() *fileState {
 	}
 }
 
+func FromLocal(ctx context.Context, filepath string, opts ...Opt) error {
```
odsod, line 58: Missing documentation here

odsod, line 58: Missing documentation here

alethenorio: Sure. Added some comments on the function

odsod (approved):
I would prefer to keep the API surface as small as possible, but if you say this is needed then let's go for it.

alethenorio:
The main use case I have seen so far is for handling unsupported archive formats (such as bz2 though I think I may have a solution for that) where there is no native Go implementation so you end up using FromRemote to download the file, then some other way to decompress the file and then FromLocal to further process the resulting tar file.

viktorvoltaire (approved):
Nice :+1:

PR #247, opened 2022-05-04
feat(protoc): keep include directory

Edholm:
The include dir is useful since it allows you to compile protobuf files that reference for example `google/protobuf/timestamp.proto`. When removed we need to manually include these from elsewhere which is a bit of a hassle.

Also upgrade `protoc` to the latest version.

The include dir must be adjacent to the binary for protoc to pick up it btw, so can't really be moved elsewhere without altering the invocation flags.

### DIFF
--- a/tools/sgprotoc/tools.go
+++ b/tools/sgprotoc/tools.go
@@ -3,7 +3,6 @@ package sgprotoc
 import (
 	"context"
 	"fmt"
-	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


Edholm:
@viktorvoltaire @odsod do you know the reason why the include dir was removed? Any other reason than "keeping it tidy"? :thinking:

Edholm:
This conflicts with buf so closing.

PR #253, opened 2022-05-31
fix(sggit): include untracked files in verify no diff

ericwenn:
Currently `VerifyNoDiff` will not fail if the staging area
is dirty, but only due to untracked files.

For example if the build process creates a new file or directory
which is not checked in, this function will not complain.


thall (approved):
:star:

ericwenn:
@odsod @thall included output regarding which files are dirty (even if they are not tracked)

odsod (approved):
Very good!

PR #259, opened 2022-07-06
feat(settings): bootstrap with inital settings

felixmr1:
I have left all the default entries, but have commented out stuff I feel is unessecary. And changed other ones to more einride-esq values (eg rebase merge over squash merge). 

Lets have some discussions around this one!

### DIFF
--- a/example/.github/settings.yml
+++ b/example/.github/settings.yml
@@ -1,0 +1,142 @@
+# These settings are synced to GitHub by https://probot.github.io/apps/settings/
+
+repository:
+  # See https://docs.github.com/en/rest/reference/repos#update-a-repository for all available settings.
+
+  # The name of the repository. Changing this will rename the repository
+  #name: repo-name
+
+  # A short description of the repository that will show up on GitHub
+  #description: description of repo
+
+  # A URL with more information about the repository
+  #homepage: https://example.github.io/
+
+  # A comma-separated list of topics to set on the repository
+  #topics: github, probot
+
+  # Either `true` to make the repository private, or `false` to make it public.
+  private: true
+
+  # Either `true` to enable issues for this repository, `false` to disable them.
+  has_issues: true
+
+  # Either `true` to enable projects for this repository, or `false` to disable them.
+  # If projects are disabled for the organization, passing `true` will cause an API error.
+  has_projects: false
+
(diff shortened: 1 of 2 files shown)


alethenorio:
Would be great to get this under source control but does this means the application would overwrite whatever was done when the repo was created?

Since this would also require admin to the repo (maybe even the org) we should check if the tool can be trusted to do so

odsod (commented):
Is this a GitHub-supported feature, or just developed by engineers who work at GitHub?

felixmr1:
> Would be great to get this under source control but does this means the application would overwrite whatever was done when the repo was created?
> 
> 
> 
> Since this would also require admin to the repo (maybe even the org) we should check if the tool can be trusted to do so


Yes I think it will. I guess one would boostrap with a minimal config and make changes that is appropriate to the repo.

felixmr1:
> Is this a GitHub-supported feature, or just developed by engineers who work at GitHub?

good question

https://github.com/apps/settings

alethenorio:
Looks like it is not GitHub

Also saw this in the notes

```
WARNING: Note that this app inherently escalates anyone with push permissions to the admin role, since they can push config settings to the master branch, which will be synced
```

Edholm:
If we go for something like this, would it be more suitable to put in our template repos?

maxekman:
Eventually I think some kind of Einride repo-operator tool that constantly analyzes our repos and make suggestions in PRs is a good solution. It gives us all the power, and that operator tool can have privileges to also do other repo setting changes from an org-level.

felixmr1:
> Looks like it is not GitHub
> 
> Also saw this in the notes
> 
> ```
> WARNING: Note that this app inherently escalates anyone with push permissions to the admin role, since they can push config settings to the master branch, which will be synced
> ```

saw that just now as well! that is a game breaker for this solution, imo!

alethenorio:
It might not be the end of the world if the repo is set to require PR reviews but definitely gives something to think about

odsod:
Let's put this on pause until next quarter and see what goes into the roadmap of @alethenorio's team?

alethenorio:
Great idea. We can evaluate this soon as part of better automation across our repos

PR #262, opened 2022-07-12
feat(deps): bump convco to 0.3.11

odsod:


Edholm (approved):
Seems to support formatting now. That's neat!

Edholm:
Should we merge to include it in next Dependabot cycle? :)

PR #268, opened 2022-07-18
feat(sgcloudspanner): add RunEmulator helper

odsod:
Starts a dockerized Cloud Spanner emulator and sets the
SPANNER_EMULATOR_HOST environment variable.

### DIFF
--- a/tools/sgcloudspanner/emulator.go
+++ b/tools/sgcloudspanner/emulator.go
@@ -1,0 +1,157 @@
+package sgcloudspanner
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"net"
+	"os"
+	"os/exec"
+	"strings"
+	"time"
+
+	"go.einride.tech/sage/sg"
+)
+
+// RunEmulator runs the Cloud Spanner emulator in Docker.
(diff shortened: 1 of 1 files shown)


Edholm (approved):
Nice! Excellent as always :godmode:

tools/sgcloudspanner/emulator.go
```diff
@@ -0,0 +1,161 @@
...
+	} else {
+		dockerRunCmd = sg.Command(ctx, "docker", "run", "-d", "--publish-all", image)
+	}
+	var dockerRunStdout strings.Builder
+	dockerRunCmd.Stdout = &dockerRunStdout
+	if err := dockerRunCmd.Run(); err != nil {
+		return nil, err
+	}
+	containerID := strings.TrimSpace(dockerRunStdout.String())
+	cleanup := func() {
```
sfinnman, line 53: Could we defer cleanups and do an if check on the err named return?

odsod: I think the readability benefit would be marginal - would prefer to keep as is

sfinnman: Alright! Still a little green on go conventions!

sfinnman (approved):
LGTM

PR #279, opened 2022-07-24
feat(sgcloudendpoints): improve helpers

odsod:
Split the DockerImage helper into the more composable DeployConfig and
BuildImage helpers.

No longer assume that the Artifact Registry is called `docker`.

### DIFF
--- a/tools/sgcloudendpoints/deprecated.go
+++ b/tools/sgcloudendpoints/deprecated.go
@@ -1,0 +1,100 @@
+package sgcloudendpoints
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"os"
+	"path/filepath"
+
+	"go.einride.tech/sage/sg"
+	"go.einride.tech/sage/tools/sgdocker"
+	"go.einride.tech/sage/tools/sggcloud"
+)
+
+// ConfigID returns the ID of the latest config for the provided Cloud Endpoints service.
+// Deprecated: Use sgcloudendpoints.DeployConfig instead.
+func ConfigID(ctx context.Context, serviceName, gcpProject string) string {
+	sg.Logger(ctx).Printf(
+		"retrieving endpoints configID from %s in %s...",
+		serviceName,
+		gcpProject,
+	)
+	return sg.Output(sggcloud.Command(
+		ctx,
+		"endpoints",
+		"configs",
+		"list",
+		"--service",
+		serviceName,
+		"--project",
+		gcpProject,
+		"--limit",
+		"1",
+		"--format",
+		"value(id)",
+	))
+}
+
+// DockerImage builds a Cloud Endpoints Docker image.
+// Deprecated: Use sgcloudendpoints.DeployConfig and sgcloudendpoints.BuildImage instead.
+func DockerImage(ctx context.Context, serviceName, gcpProject, gcpRegion string) string {
+	configID := ConfigID(ctx, serviceName, gcpProject)
+	sg.Logger(ctx).Printf(
+		"building image for %s with configID %s...",
+		serviceName,
+		configID,
+	)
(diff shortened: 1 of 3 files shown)


tools/sgcloudendpoints/tools.go
```diff
@@ -4,100 +4,159 @@ import (
...
+
+// GetServiceConfig fetches a full service config from Cloud Endpoints.
+func GetServiceConfig(ctx context.Context, service, configID string) (string, error) {
+	var accessTokenOutput strings.Builder
+	cmd := sggcloud.Command(ctx, "auth", "print-access-token")
+	cmd.Stdout = &accessTokenOutput
+	if err := cmd.Run(); err != nil {
+		return "", err
+	}
+	request, err := http.NewRequestWithContext(
```
alethenorio, line 136: Isn't this simpler to do with gcloud directly. Something like

```
gcloud endpoints configs describe --service=my-service <Config ID>
```

tools/sgcloudendpoints/deprecated.go
```diff
@@ -0,0 +1,100 @@
...
+	"os"
+	"path/filepath"
+
+	"go.einride.tech/sage/sg"
+	"go.einride.tech/sage/tools/sgdocker"
+	"go.einride.tech/sage/tools/sggcloud"
+)
+
+// ConfigID returns the ID of the latest config for the provided Cloud Endpoints service.
+// Deprecated: Use sgcloudendpoints.DeployConfig instead.
```
alethenorio, line 16: `DeployConfig` requires a config to be passed. I don't see a function to substitute being able to get what the latest config is, is that not needed?

alethenorio (approved):
Not sure the new API replaces the deprecated one 1/1 but maybe we are not using it anywhere?

tools/sgcloudendpoints/deprecated.go
odsod: Instead of querying for the latest config - we can use the return value from `DeployConfig.

tools/sgcloudendpoints/tools.go
odsod: Hmm! That sounds right! Let me try it

tools/sgcloudendpoints/deprecated.go
alethenorio: Yes but that requires deploying something while the user might just want to inquire the latest config but without looking deeper into current usages I do get a feeling that the old API is likely being used immediately after deploying anyway and for this use-case this is a better API

tools/sgcloudendpoints/tools.go
odsod: @alethenorio Updated

tools/sgcloudendpoints/deprecated.go
odsod: Yes! I'm pushing a POC PR after we merge this one - we use it immediately after deploying to build the container image

tools/sgcloudendpoints/tools.go
odsod: @alethenorio I tested it out and it seems that the output is not 100% equivalent (the service refuses to start when this config is embedded into the container).

I think it has to do with the `view=FULL` parameter which is not exposed via the CLI.

alethenorio: Ah that's too bad. Fetching through http works just fine as well

odsod: Let's go with the HTTP fetch for now? We got it from here, so there's probably intent behind it: https://github.com/GoogleCloudPlatform/esp-v2/blob/master/docker/serverless/gcloud_build_image

alethenorio: Yes let's do that

PR #280, opened 2022-07-24
feat: increase portability by bootstrapping Go

odsod:
This enables Sage to run in build environments where Go is not installed
by default - for example the `node` image.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,18 +5,37 @@
 
 sagefile := .sage/bin/sagefile
 
+# Setup Go.
+go := $(shell command -v go 2>/dev/null)
+ifndef go
+SAGE_GO_VERSION ?= 1.18.4
+export GOROOT := .sage/tools/go/$(SAGE_GO_VERSION)/go
+export PATH := $(PATH):$(GOROOT)/bin
+go := $(GOROOT)/bin/go
+os := $(shell uname | tr '[:upper:]' '[:lower:]')
+arch := $(shell uname -m)
+ifeq ($(arch),x86_64)
+arch := amd64
+endif
+$(go):
+	$(info installing Go $(SAGE_GO_VERSION)...)
+	@mkdir -p $(dir $(GOROOT))
+	@curl -sSL https://go.dev/dl/go$(SAGE_GO_VERSION).$(os)-$(arch).tar.gz | tar xz -C $(dir $(GOROOT))
+	@touch $(GOROOT)/go.mod
+	@chmod +x $(go)
+endif
+
 .PHONY: $(sagefile)
-$(sagefile):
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -5,18 +5,37 @@
 
 sagefile := .sage/bin/sagefile
 
+# Setup Go.
+go := $(shell command -v go 2>/dev/null)
+ifndef go
+SAGE_GO_VERSION ?= 1.18.4
+export GOROOT := .sage/tools/go/$(SAGE_GO_VERSION)/go
+export PATH := $(PATH):$(GOROOT)/bin
+go := $(GOROOT)/bin/go
```
alethenorio, line 14: Is this necessary? We are always adding bin to the PATH.

sg/makefile.go
```diff
@@ -57,23 +59,37 @@ func generateMakefile(ctx context.Context, g *codegen.File, pkg *doc.Package, mk
...
 	g.P(".PHONY: sage")
 	g.P("sage:")
-	g.P("\t@git clean -fxq $(sagefile)")
 	g.P("\t@$(MAKE) $(sagefile)")
 	g.P()
 	g.P(".PHONY: update-sage")
-	g.P("update-sage:")
-	g.P("\t@cd ", includePath, " && go get -d go.einride.tech/sage@latest && go mod tidy && go run .")
+	g.P("update-sage: $(go)")
+	g.P("\t@cd ", includePath, " && $(go) get -d go.einride.tech/sage@latest && $(go) mod tidy && $(go) run .")
```
alethenorio, line 92: If we are adding the go bin folder to path, is this necessary?

alethenorio (approved):
Got a few nitpick comments but looks really good. Nice work.

Would be good to benchmark how much time the download and installation adds to a CI system

Makefile
odsod: Ah yes! This is just so that we can use it as a Make target further down - this is how we ensure that we detect if we have already downloaded Go (the `$(go)` target matches a file and the download is skipped).

sg/makefile.go
odsod: True! I'll update

odsod: Or wait now I remember why I did this - since we have `$(go)` as a dependency of the target, it felt most correct to also reference it as `$(go)` in the target itself.

odsod:
@alethenorio Re:benchmarking - yes, I'll try this out in one of our frontend repos and I'll do some measurements to see how much time this adds to the build in general. It shouldn't add more than a few seconds in the general case - and we might make those seconds up from pulling a slimmer, more standard image.

PR #284, opened 2022-07-26
fix(sgfirebasetools): return correct url when setting site parameter

backevik:
A few cases I can see happening:
- Firebase project has one site where it is named identically to the project. A few of our apps are doing this (which would make it unnecessary to specify `site` option as it would default)
- Firebase project has multiple sites but configured a default (which would make it unnecessary to specify `site` option as it would default)
- Firebase project has multiple sites with no default (which makes it necessary to specify `site`)
- Firebase project has one site where it is named differently than the project (which makes it necessary to specify `site` as we cannot deduct the used site name in an easy way)

Thoughts on this fix @odsod ?


odsod:
@backevik I'll try it out and will report back

backevik:
@odsod 

Tested in portal:

Input:
`make deploy-firebase-preview pr_number=testing`
Output:
`[deploy-firebase-preview] https://einride-dev-portal-desktop--preview-testing-yyarmpmg.web.app`

code: 
```
	sg.Deps(ctx, sg.Fn(Frontend.Build, "development"))
	channelID := "preview-" + prNumber
	sg.Logger(ctx).Printf("deploying to channel %s in %s...", channelID, firebaseProjectDev)
	url, err := sgfirebasetools.DeployPreview(ctx, sgfirebasetools.DeployPreviewOptions{
		Project:   firebaseProjectDev,
		Site:      firebaseHostingSite,
		CmdDir:    sg.FromGitRoot("frontend"),
		ChannelID: channelID,
	})
	if err != nil {
		return err
	}
	sg.Logger(ctx).Println(url)
	// only post PR comment if run via github actions.
	if os.Getenv("GITHUB_ACTIONS") != "" {
		if err := commentPreviewURL(ctx, prNumber, url); err != nil {
			return err
		}
	}
	return nil
```

Ran against my local sage environment where the change in this PR is checked out.

backevik:
@odsod ok to merge this?

odsod (approved):
Let's :ship: this!

PR #286, opened 2022-07-28
feat: add function to run spanner emulator with returned information

sfinnman:


tools/sgcloudspanner/emulator.go
```diff
@@ -60,25 +91,27 @@ func RunEmulator(ctx context.Context) (_ func(), err error) {
...
+	sg.Logger(ctx).Printf("running Cloud Spanner emulator on %s", spannerEmulator.DatabaseHost)
+	if err := awaitReachable(ctx, spannerEmulator.DatabaseHost, 100*time.Millisecond, 10*time.Second); err != nil {
 		return nil, err
 	}
-	return cleanup, nil
+	return spannerEmulator, nil
+}
+
+func (e *Emulator) Close() {
+	if e == nil || e.onClose == nil {
```
odsod, line 111: it should be possible to write the API so that these types of nil checks are not necessary - when would these happen?

tools/sgcloudspanner/emulator.go
```diff
@@ -15,17 +15,47 @@ import (
 	"go.einride.tech/sage/tools/sgdocker"
 )
 
+type Emulator struct {
+	ContainerID  string
+	DatabaseHost string
+	AdminHost    string
+	onClose      func()
```
odsod, line 22: can we avoid this stateful private field?

odsod (changes requested):

tools/sgcloudspanner/emulator.go
sfinnman: Yes, we should be able to deduce the close behavior from the containerID field on the struct.

tools/sgcloudspanner/emulator.go
sfinnman: Let me remove those, just added them for safety

tools/sgcloudspanner/emulator.go
```diff
@@ -15,17 +15,47 @@ import (
...
+	cleanup := func() {
+		emulator.Close()
+		if err := os.Unsetenv("SPANNER_EMULATOR_HOST"); err != nil {
+			sg.Logger(ctx).Printf("failed to unset SPANNER_EMULATOR_HOST: %v", err)
+		}
+	}
+	return cleanup, nil
+}
+
+func RunEmulatorContainer(ctx context.Context) (emulator *Emulator, err error) {
```
odsod, line 47: based on what the method is called, I would expect the return type to be `EmulatorContainer`?

odsod: when the private state is removed, the return value becomes a pure value and we can remove the pointer from the return signature and return by value instead

Edholm:
Still relevant?

Edholm:
@sfinnman :point_up:

PR #287, opened 2022-07-29
feat(sgfirebasetools): add option for channel expiration

ericwenn:
By default firebase sets the expiry of a channel to 7 days.
In projects with a lot of activity this can lead to the quota of
channels (50) to be hit, preventing all furhter deployments.

By setting a quicker expiration this can be avoided.

### DIFF
--- a/tools/sgfirebasetools/tool.go
+++ b/tools/sgfirebasetools/tool.go
@@ -26,6 +26,10 @@ type DeployPreviewOptions struct {
 	ChannelID string
 	// CmdDir can be set if the firebase.json file is not in root of repository.
 	CmdDir string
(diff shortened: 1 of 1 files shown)


tools/sgfirebasetools/tool.go
```diff
@@ -26,6 +26,10 @@ type DeployPreviewOptions struct {
 	ChannelID string
 	// CmdDir can be set if the firebase.json file is not in root of repository.
 	CmdDir string
+	// Expires sets the duration until channel expires.
```
thall, line 29: until deployments expire?

thall: never mind

thall (approved):

PR #298, opened 2022-08-10
fix: improve relative path handling

odsod:
Use fully qualified paths to avoid path issues when invoking from
subdirectories.


pmarkus (approved):
Schnyggt

PR #301, opened 2022-08-15
feat(sggolangcilint): possibility to run in a single directory

radhus:
The default Run() doesn't always work in monorepos with multiple go.mod
files, including in the root and in subfolders.

This changes makes it possible to specify a subdirectory to run in,
reusing the default config if possible, while keeping the existing
behavior of Run().

### DIFF
--- a/tools/sggolangcilint/tools.go
+++ b/tools/sggolangcilint/tools.go
@@ -32,6 +32,21 @@ func defaultConfigPath() string {
 	return sg.FromToolsDir(name, ".golangci.yml")
 }
 
+func CommandInDirectory(ctx context.Context, directory string, args ...string) *exec.Cmd {
+	configPath := filepath.Join(directory, ".golangci.yml")
+	if _, err := os.Lstat(configPath); errors.Is(err, os.ErrNotExist) {
+		configPath = defaultConfigPath()
+	}
+	var excludeArg []string
+	if directory == sg.FromSageDir() {
+		excludeArg = append(excludeArg, "--exclude", "(is a global variable|is unused)")
+	}
+	cmdArgs := append([]string{"run", "-c", configPath}, args...)
+	cmd := Command(ctx, append(cmdArgs, excludeArg...)...)
+	cmd.Dir = directory
+	return cmd
+}
+
 // Run GolangCI-Lint in every Go module from the root of the current git repo.
 func Run(ctx context.Context, args ...string) error {
 	var commands []*exec.Cmd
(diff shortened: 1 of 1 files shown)


tools/sggolangcilint/tools.go
```diff
@@ -28,15 +28,29 @@ func Command(ctx context.Context, args ...string) *exec.Cmd {
...
-// Run GolangCI-Lint in every Go module from the root of the current git repo.
-func Run(ctx context.Context, args ...string) error {
-	defaultConfigPath := sg.FromToolsDir("golangci-lint", ".golangci.yml")
-	if err := os.MkdirAll(filepath.Dir(defaultConfigPath), 0o755); err != nil {
-		return err
+func defaultConfigPath() string {
+	return sg.FromToolsDir(name, ".golangci.yml")
+}
+
+func CommandInModule(ctx context.Context, directory string, args ...string) *exec.Cmd {
```
alethenorio, line 35: Maybe `RunInDirectory`? What do you think?

alethenorio: Or maybe `CommandInDirectory` (I've realized that this does not return an error :sweat_smile:

radhus: Yeah, `CommandInDirectory` sounds better!

tools/sggolangcilint/tools.go
```diff
@@ -96,5 +96,14 @@ func PrepareCommand(ctx context.Context) error {
...
 		return fmt.Errorf("unable to download %s: %w", name, err)
 	}
+
+	configPath := defaultConfigPath()
+	if err := os.MkdirAll(filepath.Dir(configPath), 0o755); err != nil {
+		return err
+	}
+	if err := os.WriteFile(configPath, DefaultConfig, 0o600); err != nil {
+		return err
+	}
```
alethenorio, line 106: Can we not avoid doing this on the `PrepareCommand` which is usually used for installing the binary?

tools/sggolangcilint/tools.go
```diff
@@ -28,15 +28,29 @@ func Command(ctx context.Context, args ...string) *exec.Cmd {
...
-		return err
+	var excludeArg []string
+	if directory == sg.FromSageDir() {
+		excludeArg = append(excludeArg, "--exclude", "(is a global variable|is unused)")
 	}
+	cmdArgs := append([]string{"run", "-c", configPath}, args...)
+	cmd := Command(ctx, append(cmdArgs, excludeArg...)...)
+	cmd.Dir = directory
+	fmt.Println(cmd)
+	fmt.Println(cmd.Dir)
```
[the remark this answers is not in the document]
radhus: No, strays from printf-style debugging 🤦

tools/sggolangcilint/tools.go
radhus: I'd like the config to be present when running `CommandInDirectory`. It used to be written in `Run` every time that was called, but I felt like that was a bit excessive (and calling it every time in `CommandInDirectory` feels even more excessive).

Do you have any suggestions where to write this instead? I'd like it to be written at least often enough so that changes to it is reflected, so simply checking for existence seems not enough. 🤔

alethenorio: I think that's reasonable. Let's keep it like that

alethenorio (approved):
Just the debug prints need to be fixed otherwise all good.

radhus:
@liufuyang yes, let's do it, sorry! I was originally planning to do some more testing, but it works in my repo so I guess users bumping to this version will test it for us 😉

PR #304, opened 2022-08-18
feat(exec): add new public funcion `OutputWithErr`

felixmr1:
will also return an error, if any occurs.

### DIFF
--- a/sg/exec.go
+++ b/sg/exec.go
@@ -92,12 +92,23 @@ func hasFileReferences(line string) bool {
 // Output runs the given command, and returns all output from stdout in a neatly, trimmed manner,
 // panicking if an error occurs.
 func Output(cmd *exec.Cmd) string {
+	s, err := OutputWithErr(cmd)
+	if err != nil {
+		panic(err)
+	}
+	return s
+}
+
+// OutputWithErr runs the given command.
+// Returns the error if one occurs,
+// otherwise returns all output from stdout in a neatly, trimmed manner,.
+func OutputWithErr(cmd *exec.Cmd) (string, error) {
 	cmd.Stdout = nil
 	output, err := cmd.Output()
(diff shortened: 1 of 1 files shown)


Edholm (commented):
We want to keep the public API minimal and stable. In this case we cater to the most common use-case. If you really think this will be useful across the whole ecosystem I say let's merge, otherwise I would probably say that write this func in the repo you want to use it :)

We actually started out with this setup in the beginning but actively chose to keep it minimal.

felixmr1:
> We want to keep the public API minimal and stable. In this case we cater to the most common use-case. If you really think this will be useful across the whole ecosystem I say let's merge, otherwise I would probably say that write this func in the repo you want to use it :)
> 
> We actually started out with this setup in the beginning but actively chose to keep it minimal.

Gotcha! Yeah I rather have the error returned and let the user choose wether to panic or not.

Edholm:
> > We want to keep the public API minimal and stable. In this case we cater to the most common use-case. If you really think this will be useful across the whole ecosystem I say let's merge, otherwise I would probably say that write this func in the repo you want to use it :)
> > We actually started out with this setup in the beginning but actively chose to keep it minimal.
> 
> Gotcha! Yeah I rather have the error returned and let the user choose wether to panic or not.

I would probably argue that you can write this in your repo without almost any overhead. I.e. copy the implementation of `Output` and remove the panic :)

It is basically a convenience func to avoid having to do an error check. The `cmd.Output` func is basically what you've implemented :)

alethenorio:
Personally I find the `Output` function to almost not be needed here as it does very little compared to what you already get out of the Go stdlib and I'd rather get rid of it than introducing another one which also does very little.

felixmr1:
Okidoki, ill just implement this locally then.

PR #306, opened 2022-08-22
fix: clean path to solve crashes on Mac

radhus:
When launching sagefile with double slashes, e.g. with
`/Users/user/git/repo/subfolder//../.sage/bin/sagefile`, Go crashes when
fetching tools. Cleaning the path by calling it without double slashes
does not produce the crash.

I cannot imagine any reason why it is like this, perhaps a bug to be
filed to Go upstream. The crash reports are very long, so not including
in the commit message.


odsod (approved):

Makefile
```diff
@@ -4,7 +4,7 @@
 .DEFAULT_GOAL := default
 
 cwd := $(dir $(realpath $(firstword $(MAKEFILE_LIST))))
-sagefile := $(cwd)/.sage/bin/sagefile
+sagefile := $(realpath $(cwd)/.sage/bin/sagefile)
```
odsod, line 7: Do we need to add this below aswell, where we define the `go` variable?

radhus: I'll add it as well, and I think I was able to test it properly in a `node:latest` container.

radhus:
I filed https://github.com/golang/go/issues/54590, let's see if someone can figure out why this happens 😅

PR #312, opened 2022-09-07
feat(setupsage): bump default go ver to v1.19

ButterRangers:
I received `[go-mod-tidy] go: go.mod file indicates go 1.19, but maximum version supported by tidy is 1.18` in my github action CI after an attempted bump to golang v1.19. Time to bump the default, Y/N?

### DIFF
--- a/actions/setup/action.yml
+++ b/actions/setup/action.yml
@@ -10,7 +10,7 @@ inputs:
   go-version:
     description: The Go version to download (if necessary) and use. Supports semver spec and ranges.
     required: false
-    default: 1.18
+    default: 1.19
 
   fetch-depth:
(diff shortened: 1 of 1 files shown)


ButterRangers:
Thanks! It's what I did but I figured perhaps the default should be bumped too :)

How does the proto generation differ? I wasn't aware of this :o

liufuyang:
Yeah it is a bit of a strange and annoying thing, if you want to see more about my issue see my discussion in slack [here](https://einride.slack.com/archives/C021W4LL9RC/p1663149926445279) and I did a fix via https://github.com/einride/sage/pull/316

It just shows you those proto comments are generated differently between go 1.18 and go 1.19. Not sure if it is a go thing or a sage thing that made it happen...

PR #313, opened 2022-09-07
feat: add clean alias

liufuyang:
This should allow `make clean` to work the same way as `make clean-sage`? My assumption is that as sage is heavily glued into our build system, whenever we need to do a clean operation it will be ended up as doing `make clean-sage`. Instead of renaming `clean-sage` to `clean`, adding this alias is perhaps more backward compatible?

I am totally fine we just close this PR if we think this just brings noise but no actual benefits. As I am learning about Makefiles and sage at the moment, I guess this change does what I intend to do, and perhaps this is something nice to have?

---

I tried to test this change locally by running `go build` and using the generated `sage` file in an empty project and running it as `sage init` to generate a Makefile, but not sure why my change in the local go file is not taking any effect. Still puzzled about it...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,6 +43,9 @@ update-sage: $(go)
 clean-sage:
 	@git clean -fdx .sage/tools .sage/bin .sage/build
 
+.PHONY: clean
+clean: clean-sage
+
 .PHONY: convco-check
 convco-check: $(sagefile)
 	@$(sagefile) ConvcoCheck
--- a/sg/makefile.go
+++ b/sg/makefile.go
@@ -103,6 +103,9 @@ func generateMakefile(ctx context.Context, g *codegen.File, pkg *doc.Package, mk
 		" ",
 		filepath.Join(includePath, buildDir),
 	)
+	g.P()
+	g.P(".PHONY: clean")
+	g.P("clean: clean-sage")
 	forEachTargetFunction(pkg, func(function *doc.Func, namespace *doc.Type) {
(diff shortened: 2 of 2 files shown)


Edholm (commented):
Thanks for the PR!

This change would conflict with a `clean` target that could exist in user repos and it would also "break" dependabot updates of Sage (it would require manual intervention due to the makefiles changing). 

So with that said I don't think there is enough benefit to merge this change.

liufuyang:
Thank you, that is good to know. I will close it then. But I have a few questions perhaps you can help as well? @Edholm 

Just for learning here: How to create a `clean` target that could exist in user repos? Can one edit the `Makefile` in some repo that uses sage? I saw stuff like these written on the top of the generated `Makefile` and thought users are not supposed to change anything in this Makefile?
```
# Code generated by go.einride.tech/sage. DO NOT EDIT.
```
(First line of a typical Makefile generated by sage.)

Edholm:
> 

You would add a function signature like this to your `sagefile.go` in your repo:
`func Clean(ctx contex.Context) error`.

Signature that takes context, returns an error, and is exporter (i.e. starts with an upper-case letter) will be picked up and exported to the Makefile automatically :)

liufuyang:
Woo, this feels like some magic here, apparently, I need to learn more about the basic concept of Sage, and perhaps Golang as well. Feels like I am missing some key understanding here :P 

Thanks a lot, this definitely lights me up in some correct direction.

PR #318, opened 2022-09-15
fix(sggolangcilint): turn off --path-prefix option for now

liufuyang:
I am having some similar issues as mentioned here:
https://github.com/golangci/golangci-lint/issues/3195

```
[go-lint] 
.sage/sagefile.go:5: File is not `gci`-ed with --skip-generated -s standard,default (gci)
"go.einride.tech/sage/sgtool"
.sage/sagefile.go:7: File is not `gci`-ed with --skip-generated -s standard,default (gci)
"go.einride.tech/sage/sg"
[go-lint] level=error msg="Failed to fix issues in file .sage/sagefile.go: failed to get file bytes for .sage/sagefile.go: can't read file .sage/sagefile.go: open .sage/sagefile.go: no such file or directory"
[go-lint] exit status 1
make: *** [default] Error 1
```

not sure how wide the impact is as this issue sometimes comes to me and sometimes does not.

If it only happening on me, then I can perhaps survive by using a local sage repo.

### DIFF
--- a/tools/sggolangcilint/tools.go
+++ b/tools/sggolangcilint/tools.go
@@ -49,15 +49,19 @@ func Run(ctx context.Context, args ...string) error {
 		if _, err := os.Lstat(configPath); errors.Is(err, os.ErrNotExist) {
 			configPath = defaultConfigPath
 		}
-		pathPrefix, err := filepath.Rel(sg.FromGitRoot(), filepath.Dir(path))
-		if err != nil {
-			return err
-		}
+		// Temporarily commenting usage of --path-prefix out below to solve an issue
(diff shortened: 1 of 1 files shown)


liufuyang:
Seems the way to create this issue is to update the content in a `sagefile.go` while using sage. Then next `make` gives this problem. I am on go1.19 by the way.

radhus:
@liufuyang I had forgotten about my #301 , but that removes `--path-prefix` and changes working directory instead. Could that fix it? I'll try to remember to take time and rebase and merge that 😅

liufuyang:
closing as https://github.com/einride/sage/pull/301 solves it along the way.

PR #322, opened 2022-09-23
feat(yamlfmt): add `google/yamlfmt` over `einride/yamlfmt`

felixmr1:
this deprecates `github.com/einride/yamlfmt` in favor of
`github.com/google/yamlfmt`

A reason for this is its support of retaining line breaks when
formatting

### DIFF
--- a/tools/sgyamlfmt/tools.go
+++ b/tools/sgyamlfmt/tools.go
@@ -2,6 +2,8 @@ package sgyamlfmt
 
 import (
 	"context"
+	_ "embed"
+	"errors"
 	"fmt"
 	"os"
 	"os/exec"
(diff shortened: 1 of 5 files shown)


tools/sgyamlfmt/.yamlfmt
```diff
@@ -0,0 +1,3 @@
+formatter:
+  type: basic
+  retain_line_breaks: true
```
felixmr1, line 3: I added this as a default config. Any thoughts?

odsod: Good idea! Let's call it `yamlfmt.yaml` so that we get syntax highlighting

felixmr1: ah, lets!

odsod (commented):
Please add it to the example file also, so that we get it in new projects

tools/sgyamlfmt/.yamlfmt.yaml
```diff
@@ -0,0 +1,3 @@
+formatter:
```
odsod, line 1: You can remove the dot from the name here to unhide it

tools/sgyamlfmt/tools.go
```diff
@@ -12,23 +14,51 @@ import (
...
 func Command(ctx context.Context, args ...string) *exec.Cmd {
 	sg.Deps(ctx, PrepareCommand)
-	return sg.Command(ctx, sg.FromBinDir("yamlfmt"), args...)
+	return sg.Command(ctx, sg.FromBinDir(name), args...)
+}
+
+// Run, runs the google/yamlfmt tool.
+// If no config file is found in the git root then it will run with the default config.
+func Run(ctx context.Context, args ...string) error {
+	defaultConfigPath := sg.FromToolsDir(name, defaultConfigName)
```
odsod, line 34: You can still write the file using the default name '.yamlfmt'

felixmr1: Ah, understood what ur getting at. Keep a "sane" file name here, whilst writing it as `yamlfmt` expects it.

odsod: Exactly!

odsod (approved):

PR #323, opened 2022-09-27
chore(convco): update to version 0.3.12

liufuyang:
Update to new version https://github.com/convco/convco/releases/tag/v0.3.12

`--ignore-reverts` is introduced with the new version and if people want they can turn it on for their backend service so those Github `Revert` button generated message will be ignored by convco check.

Noticing a side effect is the changelogs will also miss this `Revert "some last commit message"` as well. But I guess for backend services the changelogs are not that important so perhaps it is fine.

### DIFF
--- a/tools/sgconvco/tools.go
+++ b/tools/sgconvco/tools.go
@@ -15,7 +15,7 @@ import (
 
 const (
 	name    = "convco"
-	version = "0.3.8"
+	version = "0.3.12"
 )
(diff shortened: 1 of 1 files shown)


Edholm (approved):
Nice work!

liufuyang:
Thank you, let's give it a try later and see if other places stops me merging reverts :P

PR #326, opened 2022-09-30
feat(actions): add input to checkout different branch than default

felixmr1:


felixmr1:
This should work: https://github.com/einride/simian-adapter/actions/runs/3159370057/jobs/5142518679

image

actions/setup/action.yml
```diff
@@ -16,6 +16,10 @@ inputs:
     description: Number of commits to fetch. 0 indicates all history for all branches and tags.
     default: 0
 
+  git-branch:
+    description: What branch to checkout
+    default: master
```
felixmr1, line 21: I guess this introduces a hard dependency on master as default branch (if that requirement did not exist before).

felixmr1:
After walking this through with @Edholm we said that this might just introduce bloat, for a specifc usecase. So if anyone else dont see the imidiate need for this, I can just close it.

actions/setup/action.yml
odsod: That's OK IMO - we should have this standardized

alethenorio:
@felixmr1 what problem is this solving. Not sure I understand

felixmr1:
> @felixmr1 what problem is this solving? Not sure I understand

We sometimes have very long living branches. In this case it would enable us to have a github action schedule trigger on those.

alethenorio:
Isn't that how it already works today? If you open a PR on a branch won't it do a checkout on the branch of that PR? And any commits you push to that PR will trigger checkouts of that same branch (or commit) until you merge it?

felixmr1:
> Isn't that how it already works today? If you open a PR on a branch won't it do a checkout on the branch of that PR? And any commits you push to that PR will trigger checkouts of that same branch (or commit) until you merge it?

Yes exactly, thats how it works when a github action has been triggerd with either a pull request or a branch push. It is not the same if it gets triggered via a scheduled run.

felixmr1:
@maxekman Mentioned that we still might want this in AET. Is that correct? :)

Edholm:
@felixmr1 have you found an additional use for this? Otherwise I vote we close it.

felixmr1:
> @felixmr1 have you found an additional use for this? Otherwise I vote we close it.

Sure. Even though I will have nightmares supporting this :cry: 
```
name: Selmer
on:
  schedule:
    - cron: "0 0 * * *"

  workflow_dispatch:

jobs:
  nightly:
    runs-on: ubuntu-latest

    permissions:
      contents: write
      id-token: write
      packages: write

    env:
      GIT_BRANCH: selmer-release-candidates

    steps:
      - name: Checkout
        uses: actions/checkout@v3
        with:
          # Needed for conventional commit linting.
          fetch-depth: 0
          ref: ${{ env.GIT_BRANCH }}

      - name: Set up Go
        uses: actions/setup-go@v3
        with:
          go-version: 1.19

      - name: Cache Go modules
        uses: actions/cache@v3
        with:
          path: |
            ~/.cache/go-build
            ~/go/pkg/mod
          key: ${{ runner.os }}-go-${{ github.workflow }}.${{ github.job }}-1.19-${{ hashFiles('**/go.sum') }}
          restore-keys: |
            ${{ runner.os }}-go-${{ github.workflow }}.${{ github.job }}-1.19
      - name: Cache sage folders
        uses: actions/cache@v3
        with:
          path: |
            ./.sage/tools
            ./.sage/bin
          key: ${{ runner.os }}-sage-${{ github.workflow }}.${{ github.job }}-1.19-${{ hashFiles('./.sage/go.sum') }}
          restore-keys: |
            ${{ runner.os }}-sage-${{ github.workflow }}.${{ github.job }}-1.19
      - name: Configure git for private modules
        env:
          TOKEN: ${{ secrets.EINRIDEBOT_PERSONAL_ACCESS_TOKEN }}
        run: git config --global url."https://git:${TOKEN}@github.com/einride/".insteadOf "https://github.com/einride/"

      - name: Log in to GitHub Container registry
        uses: docker/login-action@v2.1.0
        with:
          registry: ghcr.io
          username: ${{ github.actor }}
          password: ${{ secrets.GITHUB_TOKEN }}

      - name: Trigger Orbis
        run: |
          make trigger-nightly-orbis \
            auth_token=${{ github.actor == 'dependabot[bot]' && secrets.DEPENDABOT_ORBIS_PAT || secrets.ORBIS_AUTH_TOKEN }} \
            git_branch=${{ env.GIT_BRANCH }} \
            git_sha=${{ github.sha }} \
            git_repo=${{ github.repository }}
        env:
          GITHUB_TOKEN: ${{ secrets.EINRIDEBOT_PERSONAL_ACCESS_TOKEN }}
```

Edholm (approved):
This would alleviate some pain for Felix and team so let's merge :+1:

radhus:
This PR now breaks repos which has a non-master default branch (e.g. `main`) and doesn't yet specify this tag. I wonder if a default setting could be `""` instead. WDYT?
https://github.com/einride/saga-auth/actions/runs/3370525804/jobs/5591637486

felixmr1:
> This PR now breaks repos which has a non-master default branch (e.g. `main`) and doesn't yet specify this tag. I wonder if a default setting could be `""` instead. WDYT? https://github.com/einride/saga-auth/actions/runs/3370525804/jobs/5591637486

I highlighted this here:
image

felixmr1:
@radhus so I guess you should update to use master?

radhus:
> I highlighted this here: 

Ah I missed that resolved comment. It might be our bad as we're using `einride/sage/actions/setup@master` (and not a pinned version) 😅 Not sure what is recommended here, people typically using a pinned version of the action?

> @radhus so I guess you should update to use master?

Could do that, but perhaps we should document it. It is a public tool so our standards don't apply globally 😅

PR #327, opened 2022-10-04
fix(golangcilint): disable auto fixing

radhus:
There seem to be race conditions between linters and fixers when running in parallel.

Prefer finding as many lint errors as possible, as opposed to convenience of auto fixing.


alethenorio (approved):

nils-janson (approved):

Edholm (approved):
How unfortunate :(

PR #328, opened 2022-10-07
feat: bubble up `check-latest` option from setup-go

DavyJ0nes:
Within the github workflows we want to not need to specify the patch
version of Go and always get the latest. Unfortunately cache
invalidation is a hard problem so this will bubble up the check-latest
flag from the setup-go action so that this can be specified optionally
for services.

### DIFF
--- a/actions/setup/action.yml
+++ b/actions/setup/action.yml
@@ -12,6 +12,11 @@ inputs:
     required: false
     default: 1.18
 
+  check-latest:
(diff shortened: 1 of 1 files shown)


odsod (approved):
LGTM @Edholm had you also looked at this?

Edholm (approved):
Yass, LGTM also! :100:

PR #329, opened 2022-10-09
bump deps

odsod:
- feat(api-linter): bump to 1.36.2
- feat(golangci-lint): bump to 1.50.0
- feat(goreleaser): bump to 1.11.5

### DIFF
--- a/internal/codegen/file.go
+++ b/internal/codegen/file.go
@@ -78,7 +78,7 @@ func (f *File) Write(p []byte) (int, error) {
 	if err != nil {
 		f.err = fmt.Errorf("write: %w", err)
 	}
-	return n, err //nolint: wrapcheck // false positive
+	return n, err //nolint:wrapcheck // false positive
 }
 
 // RawContent returns the raw content of the file.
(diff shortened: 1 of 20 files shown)


alethenorio (approved):

odsod:
@alethenorio The build failure looks like an issue with the `gci` linter - I'll have a look

alethenorio:
I think we've had a few people recently run into trouble with `gci`. Doesn't always seem to fix stuff

odsod:
It looks like several linters are tripping up on this specifically (gci, gofumpt)
image

alethenorio:
The correct way as per golangci lint is no space `//nolint` so maybe we need to fix our code?

Edholm:
I think the issue is that the bumped GCI inserts space, while other linters complains about the newly inserted space :(

odsod:
I fixed it @Edholm @alethenorio - the issue was that nolint comments should not have a space after the `:` - that's what all the linters were going on about! we will have some lint errors from this, but now we are enforcing the correct formatting

PR #336, opened 2022-10-13
feat: infer correct Go module path for Sage module

odsod:
Having the correct module prefix is important for exclusion from
license checking.

### DIFF
--- a/main.go
+++ b/main.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"context"
 	_ "embed"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"os"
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -126,3 +130,28 @@ func addToDependabot() error {
...
 	return os.WriteFile(dependabotYMLPath, appendSageDependabotConfig(dependabotYML), 0o600)
 }
+
+func resolveSageModulePath(ctx context.Context) (string, error) {
+	const moduleName = ".sage"
+	if _, err := os.Stat("go.mod"); err != nil {
+		return moduleName, nil
+	}
+	var out bytes.Buffer
+	cmd := sg.Command(ctx, "go", "mod", "edit", "-json")
```
alethenorio, line 140: Can we avoid calling go directly and use the official package for handling gomod files?

https://pkg.go.dev/golang.org/x/mod@v0.5.1/modfile

odsod: Since it's in `x/mod` I'd rather not add it as a dependency

alethenorio: I think that should be an OK exception since the `golang.org/x` is an official repository from the Go developers. WDYT?

odsod: The point is to not have any external dependencies at all in core Sage, so I don't think the problem is who publishes the module, but to keep the modfile looking like this:

```
module go.einride.tech/sage

go 1.17
```

alethenorio: That's a nice goal but not something we strive prevent at all costs if the alternative is better but we can take that discussion outside this PR

main.go
```diff
@@ -126,3 +130,28 @@ func addToDependabot() error {
...
+			Path string
+		}
+	}
+	if err := json.Unmarshal(out.Bytes(), &modFile); err != nil {
+		return "", err
+	}
+	if modFile.Module.Path == "" {
+		return moduleName, nil
+	}
+	return modFile.Module.Path + "/" + moduleName, nil
```
alethenorio, line 156:
```suggestion
	return filepath.Join(modFile.Module.Path, moduleName), nil
```

alethenorio (approved):

PR #340, opened 2022-10-20
feat: panic if dependency cycle is found

radhus:
Saves a list of dependencies in the context and checks whether the function to be executed is in that list already. Very trivial cycle check!

I accidentally created a target that had a circular dependency, and it ended up with a deadlock which I couldn't understand for quite a while. This change makes the same mistake cause a panic instead.

Example sagefile:
```go
func Thing(ctx context.Context) error {
        sg.Deps(ctx, Another)
        return sg.Command(ctx, "echo", "thing").Run()
}

func Another(ctx context.Context) error {
        sg.Deps(ctx, Thing)
        return sg.Command(ctx, "echo", "another").Run()
}
```

Call example:
```
$ make thing
[thing] dependency cycle calling main.Another! chain: main.Another,main.Thing
make: *** [Makefile:92: thing] Error 1
```

### DIFF
--- a/sg/deps.go
+++ b/sg/deps.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"reflect"
 	"strconv"
+	"strings"
 	"sync"
 
 	"go.einride.tech/sage/sg/internal/runner"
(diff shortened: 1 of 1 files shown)


sg/deps.go
```diff
@@ -22,6 +22,13 @@ func Deps(ctx context.Context, functions ...interface{}) {
 	var wg sync.WaitGroup
 	for i, f := range checkedFunctions {
 		i, f := i, f
+		dependencies := getDependencies(ctx)
+		for _, dependency := range dependencies {
+			if dependency.ID() == f.ID() {
+				panic("dependency cycle!")
```
stenehall, line 28: Will it print what dependency causes the cycle here? Not sure how panic works with stack traces for go

radhus: It might be deductible from the stack trace, but it would make sense to include the names of dependencies in the chain in the error string, as well as the wanted function. I'll add that 👍

radhus: I realized that I don't have the entry function in the dependency list, so the output could be a bit weird. But at least it prints a hint of where the dependency cycle is 😅 See updated commit message and PR description.

thall (approved):
Nice! :+1:

sg/deps.go
```diff
@@ -22,6 +23,20 @@ func Deps(ctx context.Context, functions ...interface{}) {
...
 	for i, f := range checkedFunctions {
 		i, f := i, f
+
+		dependencies := getDependencies(ctx)
+		for _, dependency := range dependencies {
+			if dependency.ID() == f.ID() {
+				depNames := make([]string, 0, len(dependencies))
+				for _, dependency := range dependencies {
+					depNames = append(depNames, dependency.Name())
+				}
```
Edholm, line 33:
```suggestion
				depNames := make([]string, len(dependencies))
				for i, dependency := range dependencies {
					depNames[i] = dependency.Name()
				}
```

Dunno if you can do that but is usually a good idea when you want to "copy" a slice :)

Edholm (approved):
Nice addition, thanks!

sg/deps.go
radhus: Ah, sure, I'll change! Not at all happy with the code here, loop in a loop etc, but the inner loop is only executed when panicking later so maybe it's fine 😅

PR #344, opened 2022-10-26
feat(sgcloudspanner): support offline mode

radhus:
Set SPANNER_EMULATOR_SKIP_PULL to avoid pulling the spanner image before launching. This makes it possible to run the tests without internet connectivity.

This is identical to the PR in spanner-aip-go:
https://github.com/einride/spanner-aip-go/pull/277


philiphassel (approved):
Let's use the current naming convention, but maybe `ENABLE_OFFLINE_MODE` would've been a more generic variable that could've been used for not doing other unnecessary network requests :)

alethenorio (approved):

alethenorio:
Agree that an offline mode long term might be interesting (though not sure how easy would it be to maintain). Let's go with this for now

PR #345, opened 2022-10-28
feat(java): add adoptium eclipse temurin 17

kristofferwanglund:


Edholm:
Is this something you feel will be useful to a "larger" audience? If only specific for one or two repos I would suggest adding it locally to those instead.

tools/sggit/tools.go
```diff
@@ -26,7 +26,7 @@ func VerifyNoDiff(ctx context.Context) error {
 	if status.String() != "" {
 		dirtyDetails := status.String()
 		// attempt to give a nice patch output if the dirty files are tracked by git
-		if patchOutput := sg.Output(Command(ctx, "diff", "--patch")); patchOutput != "" {
+		if patchOutput := sg.Output(Command(ctx, "diff", "--name-only")); patchOutput != "" {
```
Edholm, line 29: This might break things in unexpected ways so let's not include in this PR :)

kristofferwanglund: Good idea, it was just something I have thought of.

kristofferwanglund:
@Edholm good question. It is for one specific repo so I could add it there. However, Sage already has maven support and maven requires Java so I think this change makes Sage more complete.

tools/sggit/tools.go
kristofferwanglund: Fixed now.

PR #346, opened 2022-10-28
feat(sgphrase): add command to download phrase translations

stenehall:


tools/sgphrase/tools.go
```diff
@@ -0,0 +1,41 @@
...
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+
+	"go.einride.tech/sage/sg"
+)
+
+func Run(
```
ericwenn, line 13: How about calling this `PullTranslations`?

ericwenn (approved):

tools/sgphrase/tools.go
stenehall: Sure.

Edholm:
Is this something you feel will be useful to many repos or just one or two?

Maybe also add a comment on the pull translations func saying what it does so it will be easier for the uninitiated to know what it does :)

stenehall:
> Is this something you feel will be useful to many repos or just one or two?
> 
> Maybe also add a comment on the pull translations func saying what it does so it will be easier for the uninitiated to know what it does :)

The idea is for this to be used by all frontends in Saga moving forward. I'll add a comment.

stenehall:
Sorry for the confusion around this. I first wanted to avoid using Phrase CLI to not pull in more dependencies but it was really not the right way to go. I updated this PR with just pulling their CLI that allows us to handle translations much better.

Edholm (approved):
Nice, I think this much better aligns with how other tooling are setup :+1:

PR #347, opened 2022-11-01
fix(sgphrase): download binary instead of tar ball

philiphassel:
The binary name inside the tar ball for macosx and linux is different. Download and rename the binary instead.


stenehall:
I dislike Phrase :|

stenehall (approved):

PR #348, opened 2022-11-01
revert(actions): add input to checkout different branch than default

radhus:
This reverts commit e443bb44aea63957b34a37474e1ec44dadaa2fc2.

Specifying a branch, or defaulting to master, makes PR tests run on the latest commit on that branch, not on the PR commit.

### DIFF
--- a/actions/setup/action.yml
+++ b/actions/setup/action.yml
@@ -21,10 +21,6 @@ inputs:
     description: Number of commits to fetch. 0 indicates all history for all branches and tags.
     default: 0
 
-  git-branch:
-    description: What branch to checkout
-    default: master
-
 runs:
   using: composite
 
@@ -34,7 +30,6 @@ runs:
       with:
         # Needed for conventional commit linting.
         fetch-depth: ${{ inputs.fetch-depth }}
-        ref: ${{ inputs.git-branch }}
 
     - name: Set up Go
       uses: actions/setup-go@v3


actions/setup/action.yml
```diff
@@ -21,10 +21,6 @@ inputs:
     description: Number of commits to fetch. 0 indicates all history for all branches and tags.
     default: 0
 
-  git-branch:
-    description: What branch to checkout
-    default: master
```
radhus, line 26: Not sure if this is backwards compatible, perhaps only remove the usage of it below?

felixmr1:
So I guess what one would do is to set the ref explicitly if one wants to change, but not having it at all if you want the PR branch.

So something like:
```

  steps:
    - name: Checkout
      if: ${{ inputs.git-branch }} == ""
      uses: actions/checkout@v3
      with:
        # Needed for conventional commit linting.
        fetch-depth: ${{ inputs.fetch-depth }}

    - name: Checkout other branch
      if: ${{ inputs.git-branch }} != ""
      uses: actions/checkout@v3
      with:
        # Needed for conventional commit linting.
        fetch-depth: ${{ inputs.fetch-depth }}
        ref: ${{ inputs.git-branch }}

odsod (approved):

actions/setup/action.yml
radhus: Let's hope this hasn't been used widely yet, and deal with any breakage 🙏

PR #355, opened 2022-11-21
feat: add `enableCache` option to sage github action

philiphassel:
Can be used to disable cache.


Edholm (approved):
:+1:

philiphassel:
Should it be called `enableCache` (and default to `true`) or `disableCache` (and default to`false`)?

thall (approved):

PR #359, opened 2022-12-01
fix(spantest): use `docker port` to lookup port mapping

thall:
This also solves the problem that the host is empty if running with
Podman instead of Docker.


Edholm (approved):
Neat!

radhus (approved):
Much simpler code! 🤩

alethenorio (commented):
@thall Looking at the examples [here](https://docs.docker.com/engine/reference/commandline/port/) is this going to capture any potential line breaks as well in the output? Would it be safer to add some trimming on the output string?

thall:
@alethenorio I will update this, since it will return ipv6 mapping as well, you can see it here https://github.com/einride/spanner-aip-go/actions/runs/3590542483/jobs/6044034145

Working on a fix.

PR #361, opened 2022-12-02
fix(gha-actions): evaluate composite input bool as string

saimanwong:
https://github.com/actions/runner/issues/1483
https://github.com/actions/runner/issues/1483#issuecomment-1279933184

---

I've noticed that several CI jobs have been pretty slow recently and most repositories have empty caches (e.g., pick any repository, go to **Actions** -> **Caches**)

Example log when running the composite setup sage action,

Currently (no `actions/cache@v3` log row):

Screenshot 2022-12-02 at 15 09 37

After this change:

Screenshot 2022-12-02 at 15 13 56

### DIFF
--- a/actions/setup/action.yml
+++ b/actions/setup/action.yml
@@ -43,7 +43,7 @@ runs:
         check-latest: ${{ inputs.check-latest }}
 
     - name: Cache Go modules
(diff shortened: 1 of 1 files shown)


Edholm (approved):
Nice find!

PR #362, opened 2022-12-08
feat: add firebase config for CSP

stenehall:
Very initial take on a suggested way of handling CSP for firebase.

This takes in a set of "groups" and saves a generated Firebase config file using it. If we take the decision to use something like this we have to go over each saga app and add groups for all domains we want to add

Usage:
```go
func (f Frontend) AddFirebaseConfig(ctx context.Context) error {
	csp := sgfirebaseconfig.Init()
	csp.AddRule("einride-dev-default", sgfirebaseconfig.Group{
		ConnectSrc: []string{"wss://*.einride.dev", "https://*.einride.dev", "https://cookie.saga.einride.tech"},
		FontSrc:    []string{"https://fonts.einride.tech"},
		StyleSrc:   []string{"'self'", "'unsafe-inline'", "https://fonts.einride.tech"},
		ScriptSrc:  []string{"'self'"},
		ImgSrc:     []string{"'self'", "data:"},
		WorkerSrc:  []string{"blob:"},
	})
	err := sgfirebaseconfig.Update(
		"firebase.json",
		"einride-orchestrate-app-dev",
		[]string{"firebase-auth", "einride-dev-default", "amplitude"},
		"https://setry.foobar.com",
	)
}
```

firebase.json

```json
{
 "hosting": [
  {
   "headers": [
    {
     "headers": [
      {
       "key": "Content-Security-Policy",
       "value": "default-src 'self'; connect-src https://identitytoolkit.googleapis.com https://securetoken.googleapis.com wss://*.einride.dev https://*.einride.dev https://cookie.saga.einride.tech https://api2.amplitude.com/2/httpapi; font-src https://fonts.einride.tech; frame-src ; img-src 'self' data:; script-src 'self'; style-src 'self' 'unsafe-inline' https://fonts.einride.tech; worker-src blob:; report-uri ; report-to default"
      }
     ],
     "source": "/**"
    },
    {
     "headers": [
      {
       "key": "Cache-Control",
       "value": "max-age=120"
      }
     ],
     "source": "/**"
    },
    {
     "headers": [
      {
       "key": "Cache-Control",
       "value": "max-age=604800"
      }
     ],
     "source": "**/*.@(jpg|jpeg|gif|png|svg|webp|js|css|eot|otf|ttf|ttc|woff|woff2|font.css)"
    }
   ],
   "ignore": [
    "firebase.json",
    "**/.*",
    "**/node_modules/**"
   ],
   "public": "build",
   "rewrites": [
    {
     "destination": "/index.html",
     "source": "**"
    }
   ],
   "site": "einride-orchestrate-app-dev"
  }
 ]
}
```

### DIFF
--- a/tools/sgfirebaseconfig/tools.go
+++ b/tools/sgfirebaseconfig/tools.go
@@ -1,0 +1,214 @@
+package sgfirebaseconfig
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
+	"strings"
+
+	"go.einride.tech/sage/sg"
+)
+
+type CSP struct {
+	Rules map[string]Group
+	Ctx   context.Context
+}
+
+type Group struct {
+	ConnectSrc []string
+	FontSrc    []string
+	FrameSrc   []string
+	ImgSrc     []string
+	ScriptSrc  []string
+	StyleSrc   []string
+	WorkerSrc  []string
+}
+
+func Init(ctx context.Context) CSP {
+	var csp CSP
+	csp.Ctx = ctx
+	csp.Rules = make(map[string]Group, 0)
+	csp.AddRule("firebase-auth", Group{
+		ConnectSrc: []string{
+			"https://identitytoolkit.googleapis.com",
+			"https://securetoken.googleapis.com",
+		},
+	})
+	csp.AddRule("sentry", Group{
+		ConnectSrc: []string{"https://*.ingest.sentry.io"},
+		ScriptSrc:  []string{"https://sentry.io"},
+	})
+	csp.AddRule("amplitude", Group{
+		ConnectSrc: []string{"https://api2.amplitude.com/2/httpapi"},
+	})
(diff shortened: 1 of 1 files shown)


tools/sgfirebaseconfig/tools.go
```diff
@@ -0,0 +1,236 @@
...
+	connectSrc []string
+	fontSrc    []string
+	frameSrc   []string
+	imgSrc     []string
+	scriptSrc  []string
+	styleSrc   []string
+	workerSrc  []string
+}
+
+func (csp *CSP) rules() { // Default list of things needed for a Saga app.
```
thall, line 28: All this rules are very Einride specific. Sage is open source and I don't really see how this fits here, or do I miss something?

stenehall: Rewrite this now. Broke out the `AddRule` functionality so that we can add anything company specific in the projects sage config. Have a look at the updated PR comment above.

PR #364, opened 2022-12-16
pin spanner version

thall:
- feat(sgcloudspanner): pin Cloud Spanner Emulator version
- feat(sgcloudspanner): only pull image if missing locally


odsod (approved):
Surgical! :mag_right:

tools/sgcloudspanner/emulator.go
```diff
@@ -34,12 +39,11 @@ func RunEmulator(ctx context.Context) (_ func(), err error) {
...
-	const image = "gcr.io/cloud-spanner-emulator/emulator:latest"
 	if _, ok := os.LookupEnv("SPANNER_EMULATOR_SKIP_PULL"); !ok {
-		cmd := sgdocker.Command(ctx, "pull", image)
-		cmd.Stdout, cmd.Stderr = nil, nil
-		if err := cmd.Run(); err != nil {
-			return nil, err
+		if dockerPullNeeded(ctx) {
+			if err := dockerPull(ctx); err != nil {
+				return nil, err
+			}
```
radhus, line 46: Now that we have pinned the version (🎉 !!!), I think `docker run` will pull the image if it doesn't exist - so I think we should be able to drop all `docker pull` commands now instead 😮 WDYT?

thall: even better! :heart:

radhus (approved):
💥 💥 🧹

PR #368, opened 2023-01-02
feat: mark Makefile's as generated for GitHub

radhus:
Using the `linguist-generated` directive in .gitattributes, this tells GitHub that the Makefile's should be treated as generated files and by default not display its diff in PRs.

See documentation:
https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+Makefile linguist-generated=true
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ go run go.einride.tech/sage@latest init
 
 Run `make`.
 
-Two changes should now have happened. If the project had a previous `Makefile` it should have been renamed to `Makefile.old` and a new should have been created. If the project have a dependabot config, a sage config should have been added.
+Three changes should now have happened. If the project had a previous `Makefile` it should have been renamed to `Makefile.old` and a new should have been created. If the project have a dependabot config, a sage config should have been added. GitHub should treat Makefiles as generated code by listing it in `.gitattributes`.
 
 Usage
 -----
--- a/example/.gitattributes
+++ b/example/.gitattributes
@@ -1,0 +1,1 @@
+Makefile linguist-generated=true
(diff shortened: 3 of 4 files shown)


.gitattributes
```diff
@@ -0,0 +1 @@
+Makefile linguist-generated=true
```
odsod, line 1: Not sure this can be applied to a whole repository - how do you know that there aren't sub-directories with non-generated Makefiles?

radhus: That is a good point, didn't think of that scenario 😅 

I see two alternatives:
1. We keep this change, assume most users only have generated Makefiles, and make it more clear in documentation and `sage init` output what this implies. However that information might be forgotten when later introducing manually written Makefiles.
2. Only document suggestion to add this to your repo as a convenience.

I would prefer 1 to optimize for most users (assuming that is true), but understand if that is perhaps too bold and we should be more safe instead. WDYT?

odsod: What if we wrote absolute paths to all generated Makefiles? Feels like the only waterproof solution

radhus: Hmm, that would also make this propagate to existing repositories nicely! I'll see what I can do!

odsod: Cool! Thinking out loud here but what if we added our own attribute aswell `sage-generated=true` to be able to "clean" out attributes for removed Makefiles?

radhus: That is a good point. I don't see that we currently keep track of removed generated Makefiles anywhere, so I guess we'd have to compare the list of files-with-`sage-generated=true`-attributes with the list of generated makefiles and clean out the removed ones. 🤔 

I'll see if this is something that could easily be incorporated when I take a look at this PR again 🙏

Edholm:
@radhus should we keep this one open?

radhus:
> @radhus should we keep this one open?

I forgotten and didn't take time to complete it. I'll close it and leave it as an exercise for future readers 😬

PR #373, opened 2023-01-27
feat(firebase): limit preview deploymnet to project

stenehall:
This limits the deploy to only use the hosting:<project-name> of the provided project from the firebase.json file.

More reading about `--only` https://firebase.google.com/docs/cli/#partial_deploys

When writing this I realise that this might be a breaking change. I'm not suer how it will fallback if no project name is specified in the firebase.json file. I will have to investigate that before moving on with this. Changing this into a draft instead, for now.

### DIFF
--- a/tools/sgfirebasetools/tool.go
+++ b/tools/sgfirebasetools/tool.go
@@ -39,6 +39,7 @@ func DeployPreview(ctx context.Context, opts DeployPreviewOptions) (string, erro
 		"hosting:channel:deploy",
 		opts.ChannelID,
 		"--project", opts.Project,
+		"--only", opts.Project,
 		"--json",
 	}
 	if opts.Site != "" {


Edholm:
Maybe you could add a short description what this change entails? :)

stenehall:
> Maybe you could add a short description what this change entails? :)

Fair enough :) I was so deep in "fixing things" mode that I kind of forgot that not everyone sees the obvious here :)

backevik (commented):
Should this flag not be a firebase service instead of a project? Looking at this: https://firebase.google.com/docs/cli#partial_deploys

stenehall:
> Should this flag not be a firebase service instead of a project? Looking at this: https://firebase.google.com/docs/cli#partial_deploys

The way I read it is that you'd want to add `"--only", "hosting:"+project,`here. Doing that to `sgfirebasetools.Command` works like expected and I noticed that you're also doing that in your project. However In `DeployPreview` we already have `"hosting:channel:deploy"` as a param and that seems to change it into just requiring the actual project.

The intention for this is to make it work with multiple hosts in your `firebase.json`, something that broken when I added it.

As far as I can tell this actually works as expected for me. But would love someone else to look at it as well quickly before taking it live.

Edholm:
@stenehall should we keep this open?

PR #377, opened 2023-01-31
feat: bump sqlfluff version and add dbt templater deps

alfredgunnar:
Enables using the dbt templater with sqlfluff.


thall (approved):
LGTM :+1:

PR #391, opened 2023-02-16
feat(sggolicenses): add command for checking license in a specific path

alethenorio:
In projects where there are no Go files in the same folder as the go.mod file (often because one wants to keep the go.mod in the root of the project to keep git tagging simple) the license checker fails as it expects go source files where it is running.
So we add a new CheckInDirectory target which uses a specified directory instead of a go.mod file

### DIFF
--- a/tools/sggolicenses/tools.go
+++ b/tools/sggolicenses/tools.go
@@ -25,6 +25,38 @@ func Command(ctx context.Context, args ...string) *exec.Cmd {
 	return sg.Command(ctx, sg.FromBinDir(name), args...)
 }
 
+// Check for disallowed types of Go licenses in a specific directory.
+// By default, Google's forbidden and restricted types are disallowed.
+func CheckDir(ctx context.Context, directory string, disallowedTypes ...string) error {
+	args := []string{
+		"check",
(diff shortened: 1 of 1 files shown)


tools/sggolicenses/tools.go
```diff
@@ -25,6 +25,38 @@ func Command(ctx context.Context, args ...string) *exec.Cmd {
 	return sg.Command(ctx, sg.FromBinDir(name), args...)
 }
 
+// Check for disallowed types of Go licenses in a specific directory.
+// By default, Google's forbidden and restricted types are disallowed.
+func CheckInDirectory(ctx context.Context, directory string, disallowedTypes ...string) error {
```
odsod, line 30: since we use `Dir` in other parts of Sage (as the `sg.FromBinDir` above) let's call it `CheckDir`?

odsod (approved):

tools/sggolicenses/tools.go
alethenorio: Sure

alethenorio: Fixed it

PR #397, opened 2023-03-01
feat: add `CommandContext` to exec

Hamsajj:
It's beneficial to be able set a timeout for some commands using sage. 
~This PR adds a new `CommandContext` function to exec package to allow for that
Didn't add context to `Command` function to avoid any side affect on already available tools.~

This PR uses `exec.CommandContext` instead of `exec.Command`. 
The provided context is used to interrupt the process (by calling cmd.Cancel or os.Process.Kill) if the context becomes done before the command completes on its own ([reference](https://pkg.go.dev/os/exec#CommandContext))

### DIFF
--- a/sg/exec.go
+++ b/sg/exec.go
@@ -24,8 +24,7 @@ func ContextWithEnv(ctx context.Context, env ...string) context.Context {
 
 // Command should be used when returning exec.Cmd from tools to set opinionated standard fields.
 func Command(ctx context.Context, path string, args ...string) *exec.Cmd {
-	// TODO: use exec.CommandContext when we have determined there are no side-effects.
-	cmd := exec.Command(path)
+	cmd := exec.CommandContext(ctx, path)
 	cmd.Args = append(cmd.Args, args...)
 	cmd.Dir = FromGitRoot(".")
 	cmd.Env = os.Environ()


sg/exec.go
```diff
@@ -22,10 +22,20 @@ func ContextWithEnv(ctx context.Context, env ...string) context.Context {
...
 
+// CommandContext is like Command but uses exec.CommandContext.
+func CommandContext(ctx context.Context, path string, args ...string) *exec.Cmd {
+	cmd := exec.CommandContext(ctx, path)
+	return setStandardFields(ctx, cmd, args...)
+}
+
 // Command should be used when returning exec.Cmd from tools to set opinionated standard fields.
 func Command(ctx context.Context, path string, args ...string) *exec.Cmd {
 	// TODO: use exec.CommandContext when we have determined there are no side-effects.
```
odsod, line 33: what are your thoughts on addressing this TODO as part of this commit?

Hamsajj: At first I wanted to use context in the same `Command` function. But then thought maybe it will be a breaking change. It would not affect any existing tooling in sage if I'm not missing anything. But maybe someone is using this function with a custom workaround with timeouts and cancels on the context and changing this function will break that.

odsod: I still don't fully understand this change, but I might be looking at it from the wrong angle.

* Isn't the implementation for `Command` and `CommandContext` identical? What's the purpose of adding the new method to the API surface if it does the same thing as the old one?

* The `TODO` comment has to be addressed, since it says that we should use `exec.CommandContext` after figuring out if there are any side-effects to using it - and now with this change we are - this comment is no longer relevant?

Hamsajj: Oh. That identical implementation was a mistake, and was on me for not reviewing my own code. Fixed that. 
- `Command` will not consider `ctx` and basically does not change.
- Removed that TODO comment. I thought your initial question is why we should add another function and not call `exec.CommandContext` inside the `Command` itself.

As why I added another function instead of changing the `Command`, I have no strong opinion but as I said I thought it could be a breaking change in some edge cases.

odsod: OK now I get it, thanks!

I would rather we take a small risk for breakage (assuming that you have reasoned about what might go wrong) than add another method to the public API, especially since both methods take a context, I think it will be confusing why one method is called `Context` and another is called `CommandContext`, when both methods do accept a context.

odsod: Just post to our internal #sage channel before merging so everyone internally is aware of the change.

odsod: @Edholm @alethenorio WDYT?

Hamsajj: I've changed to use the same `Command` function. Tested it with multiple projects using sage and everything was okay.

Edholm: The only real downside I've seen using the `CommandContext` is that a `SIGKILL` is sent to the process on context cancellation. This is probably fine for most applications, but those that require cleanup (like terraform, returning the lock) will be sad.

However, given that a context cancellation (and most likely a ctrl+c) is abnormal this might be fine? Take ctrl+c for example: you would expect this to send and propagate a SIGTERM to any subprocesses, but with this change it will likely use `SIGKILL` instead. Don't take my word for that though...

Edholm: If you've considered the above, then LGTM :)

Hamsajj: tbh I did not consider SIGKILL vs SIGTERM at all. We can do something like this in sage which tries SIGTERM on context.Done() first, then retries SIGKILL after a timeout:
https://github.com/golang/go/issues/21135#issuecomment-324478980

alethenorio: Have we considered how this would propagate through `make`? As far as I understand `make` will not handle a CTRL+C and wait for the application to gracefully exit so this will only work if whatever the application is doing is exited before make does so. Also does not add any option for cleanup as @Edholm mentioned.

If we make this change, are we stuck with a breaking change if we want to look into any of this?

Hamsajj:
I did not consider how the process will be killed and differences between `SIGKILL` and `SIGTERM` . Can not merge this PR without looking more into it and testing side affects.

PR #398, opened 2023-03-02
fix: for darwin binary being compressed and couldn't run

mlveggo:


Edholm (approved):
Was it compressed without actually having a file extension indicating this? :thinking:

mlveggo:
Yes

PR #399, opened 2023-03-03
fix: upgrade dbt-core to 1.4.0

m4mm4r:
if approved, this PR will:

- Upgrade dbt core to `1.4.0`.
- Remove the temporary workaround where `pytz` is explicitly installed. Not needed anymore as this is fixed in dbt core https://github.com/dbt-labs/dbt-core/pull/7077


Edholm (approved):
Nice! Great that they fixed this upstream instead

PR #410, opened 2023-03-20
fix(sgmvn): change remote for downloading mvn

michalmac:
archive.apache.org is a long-term archive, whereas dlcdn.apache.org has a limited selection of latest
 releases. After 3.8.8 was released, 3.8.7 was removed from the latter server.


Edholm (approved):
Thanks!

PR #412, opened 2023-03-25
feat: add betteralign tool to optimize structs for memory storage

mlveggo:


odsod (approved):
:+1: Let's keep an eye on [this](https://github.com/dkorunic/betteralign/issues/1) and use it from GolangCI-Lint if it becomes available?

PR #414, opened 2023-03-27
feat: bump default Go version to 1.20.2

alethenorio:
Same version as the github action uses

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ sagefile := $(abspath $(cwd)/.sage/bin/sagefile)
 go := $(shell command -v go 2>/dev/null)
 export GOWORK ?= off
 ifndef go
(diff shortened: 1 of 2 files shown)


Edholm (approved):
Nice! Unfortunate that it will be a "breaking" change in that people will have to manually update their dependabots :D

alethenorio:
> Nice! Unfortunate that it will be a "breaking" change in that people will have to manually update their dependabots :D

How so? Maybe I missed something but shouldn't this have an effect mostly on non-go repos which will then only be used to compile sage anyway?

alethenorio:
Ah I see. You mean dependabot won't commit makefile changes to bump PRs. Yeah this has been a side effect of Sage in these kinds of changes.

Thanks for the review :heart:

PR #417, opened 2023-04-12
feat(semantic-release): add optional arguments

kristofferwanglund:
Make it possible to pass additional command line arguments
to the `semantic-release` command.

### DIFF
--- a/tools/sggosemanticrelease/tools.go
+++ b/tools/sggosemanticrelease/tools.go
@@ -23,7 +23,8 @@ func Command(ctx context.Context, args ...string) *exec.Cmd {
 	return sg.Command(ctx, sg.FromBinDir(name), args...)
 }
 
-func ReleaseFromCloudBuildCommand(ctx context.Context, ci bool, repo string) *exec.Cmd {
+// ReleaseFromCloudBuildCommand creates a semantic release with optional extra arguments to semantic-release.
+func ReleaseFromCloudBuildCommand(ctx context.Context, ci bool, repo string, extraArgs ...string) *exec.Cmd {
 	args := []string{
 		"--allow-initial-development-versions",
 		"--allow-no-changes",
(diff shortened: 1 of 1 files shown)


Edholm:
I agree with Oscar. The new function is too similar to `Command`. Could add an "OpinionatedArgs" that returns our default args? :thinking:

kristofferwanglund:
I have pushed a change where the existing command gets an optional arguments parameter. What do you think of this?

tools/sggosemanticrelease/tools.go
```diff
@@ -32,6 +33,7 @@ func ReleaseFromCloudBuildCommand(ctx context.Context, ci bool, repo string) *ex
 		"--provider-opt",
 	}
 	args = append(args, "slug="+repo)
```
kristofferwanglund, line 35: Just a thought, I think the `github-prover` requires the `slug` option, so perhaps include it in the "default args" list above?

tools/sggosemanticrelease/tools.go
```diff
@@ -23,7 +23,8 @@ func Command(ctx context.Context, args ...string) *exec.Cmd {
 	return sg.Command(ctx, sg.FromBinDir(name), args...)
 }
 
-func ReleaseFromCloudBuildCommand(ctx context.Context, ci bool, repo string) *exec.Cmd {
+// ReleaseFromCloudBuildCommand creates a semantic release with optional arguments to semantic-release.
+func ReleaseFromCloudBuildCommand(ctx context.Context, ci bool, repo string, argsOpt ...string) *exec.Cmd {
```
Edholm, line 27: Maybe call it `extraArgs`, one could otherwise argue that these should override rather than be appended I guess :shrug: 

```suggestion
func ReleaseFromCloudBuildCommand(ctx context.Context, ci bool, repo string, extraArgs ...string) *exec.Cmd {
```

tools/sggosemanticrelease/tools.go
Edholm: I don't know enough about to say

tools/sggosemanticrelease/tools.go
kristofferwanglund: Good idea, fixed.

odsod (approved):

PR #431, opened 2023-06-07
feat(sgbalena): use local installed balena binary if exists

ericwenn:
Similar to how gcloud and docker is handled; if the binary can be found
in users path, use it directly.

This makes it possible to use sgbalena on NixOS.

### DIFF
--- a/tools/sgbalenacli/tools.go
+++ b/tools/sgbalenacli/tools.go
@@ -54,6 +54,13 @@ func Whoami(ctx context.Context) (WhoamiInfo, error) {
 }
 
 func PrepareCommand(ctx context.Context) error {
+	// Special case: use local balena CLI if available.
+	if binary, err := exec.LookPath("balena"); err == nil {
+		if _, err := sgtool.CreateSymlink(binary); err != nil {
+			return err
+		}
+		return nil
+	}
 	binDir := sg.FromToolsDir(toolName, version)
 	binary := filepath.Join(binDir, toolName, binaryName)
 	hostOS := runtime.GOOS


thall (approved):

tools/sgbalenacli/tools.go
```diff
@@ -54,6 +54,13 @@ func Whoami(ctx context.Context) (WhoamiInfo, error) {
 }
 
 func PrepareCommand(ctx context.Context) error {
+	// Special case: use local balena CLI if available.
+	if binary, err := exec.LookPath("balena"); err == nil {
+		if _, err := sgtool.CreateSymlink(binary); err != nil {
+			return err
+		}
+		return nil
+	}
```
Edholm, line 63: I understand the reason for your change but I don't know how I feel about it. The reason being that:
1. There may be different behavior between your local invocation and the one in CI due to different version
2. This becomes a "hidden" feature that may end up being confusing

That said, if we use this pattern in other tools, then I'm not against it. What are your thoughts on the subject @thall @alethenorio @odsod ?

ericwenn: I'm not sure what the rationale is for the other tools (but without this behavior they wouldn't work on NixOS either).
I agree with your hesistance wrt version drift, however looking at releases for balena CLI I don't think version drift is going to be a problem in practice (not a lot of features added over time).

A workaround for NixOS is to do the symlinking manually after new balena binary is installed. If you don't like this change @Edholm I'll do that instead, but would prefer to have it automatically :)

Edholm: Yes, I feel comfortable merging. We can always revert if it becomes an issue in the future.

alethenorio: I agree with your reasoning @Edholm 

I think this can lead to weird issues you described. Given that Balena doesn't change in feature super often, let's keep an eye on this but avoid adding to more tools

PR #434, opened 2023-06-14
feat(cloudendpoints): remove deprecated functions

radhus:
These have multiple issues, including not using the fixed ESPv2 version defined.

Having been marked as deprecated for almost a year, it's time to let go.


Edholm (approved):
Great idea, thanks!

alethenorio (approved):

PR #437, opened 2023-06-16
fix(tools)!: upgreade sgcloudsqlproxy to 2.3.0 (breaking change) 

liufuyang:
Beaking change as one will need to replace the cmd parameter from
'cloud_sql_proxy -instances=myproject:myregion:myinstance=tcp:5432 --enable_iam_login'
to:
'cloud-sql-proxy myproject:myregion:myinstance --port 5432 --auto-iam-authn'

---

## Reason for update:

Perhaps this can help the build error that shows up recently from time to time?

```
[backend:deploy-dev] migrating database...
[backend:deploy-dev] 2023/06/16 08:31:09 current FDs rlimit set to 65536, wanted limit is 8500. Nothing to do here.
[backend:deploy-dev] 2023/06/16 08:31:09 using credential file for authentication; 
[backend:deploy-dev] Running golang migrate...
[backend:deploy-dev] error: dial tcp [::1]:5432: connect: connection refused
[backend:deploy-dev] could not migrate database: exit status 1
```

As this error seems happened in the db migration step and a proxy connection is needed for that.

---

The download UIL might be changed for the new version, I hope I setup it up correctly but you can help me double check, thank you
https://cloud.google.com/sql/docs/mysql/connect-auth-proxy#macos-64-bit

### DIFF
--- a/tools/sgcloudsqlproxy/tool.go
+++ b/tools/sgcloudsqlproxy/tool.go
@@ -12,8 +12,8 @@ import (
 )
 
 const (
-	version    = "1.33.1"
-	binaryName = "cloud_sql_proxy"
+	version    = "2.3.0"
+	binaryName = "cloud-sql-proxy"
 )
 
 func Command(ctx context.Context, args ...string) *exec.Cmd {
@@ -25,12 +25,12 @@ func PrepareCommand(ctx context.Context) error {
 	binDir := sg.FromToolsDir(binaryName, version)
 	binary := filepath.Join(binDir, binaryName)
 	filename := fmt.Sprintf(
-		"cloud_sql_proxy.%s.%s",
+		"cloud-sql-proxy.%s.%s",
 		runtime.GOOS,
 		runtime.GOARCH,
 	)
 	binURL := fmt.Sprintf(
-		"https://storage.googleapis.com/cloudsql-proxy/v%s/%s",
+		"https://storage.googleapis.com/cloud-sql-connectors/cloud-sql-proxy/v%s/%s",
 		version,
 		filename,
 	)


Edholm (approved):
nice thanks. Since it is a major bump, do we expect any breaking changes?

Edholm (changes requested):
I tried this locally and got: 
```
[sgcloudsqlproxy:prepare-command] fetching https://storage.googleapis.com/cloud-sql-connectors/cloudsql-proxy/v2.3.0/cloud_sql_proxy.linux.amd64 ...
[sgcloudsqlproxy:prepare-command] unable to download cloud_sql_proxy: unable to download file: download binary https://storage.googleapis.com/cloud-sql-connectors/cloudsql-proxy/v2.3.0/cloud_sql_proxy.linux.amd64: status code 404
```

I think perhaps a dash is missing in the url.

tools/sgcloudsqlproxy/tool.go
```diff
@@ -30,7 +30,7 @@ func PrepareCommand(ctx context.Context) error {
 		runtime.GOARCH,
 	)
 	binURL := fmt.Sprintf(
-		"https://storage.googleapis.com/cloudsql-proxy/v%s/%s",
+		"https://storage.googleapis.com/cloud-sql-connectors/cloudsql-proxy/v%s/%s",
```
Edholm, line 33:
```suggestion
		"https://storage.googleapis.com/cloud-sql-connectors/cloud-sql-proxy/v%s/%s",
```

filename above should change `_` to `-` also.

Edholm (commented):
This is a breaking change and the migration guide can be found here: https://github.com/GoogleCloudPlatform/cloud-sql-proxy/blob/main/migration-guide.md

Edholm:
I got it to work locally with the above changes and this change to the invocation:
image

tools/sgcloudsqlproxy/tool.go
liufuyang: aha, thanks a lot, let me update :)

liufuyang:
@Edholm Thanks again for checking the change here, I would also like to verify this on our side as well when I have time before merge, hopefully I can find time recently :)

Edholm (approved):
Feel free to merge when you've tested it locally :)

liufuyang:
Great, tested that with update as such then it works:
```go
        proxyCmd := sgcloudsqlproxy.Command(
		ctx,
		fmt.Sprintf("%s:%s:%s", gcpProject, gcpRegion, dbInstanceId),
		"--auto-iam-authn",
	)
```

PR #439, opened 2023-06-27
feat(tracing): print execution times if SAGE_TRACE=true

Edholm:
Set the SAGE_TRACE env var to true to output execution times for targets. Can be useful in CI to see what is slowing things down.

### DIFF
--- a/sg/internal/runner/runner.go
+++ b/sg/internal/runner/runner.go
@@ -2,7 +2,9 @@ package runner
 
 import (
 	"context"
+	"os"
 	"sync"
+	"time"
 )
 
 // global state for the runner.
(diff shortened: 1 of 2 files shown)


sg/internal/runner/runner.go
```diff
@@ -15,6 +17,13 @@ var (
 
 // RunOnce uses key to ensure that fn runs exactly once and always returns the error from the initial run.
 func RunOnce(ctx context.Context, key string, fn func(context.Context) error) error {
+	before := time.Now()
+	defer func() {
+		if os.Getenv("SAGE_TRACE") == "true" {
+			elapsed := time.Since(before)
+			println(key, "took", elapsed.String())
```
thall, line 24: any reason for not using the logger?

Edholm: nope, I will change if this is something we actually want merged to master :)

PR #442, opened 2023-07-06
feat: add Backstage validation helper

odsod:
Encapsulates default settings (and potential config), which we expect to
change over time.


alethenorio (approved):
:raised_hands:

PR #443, opened 2023-07-06
feat(example): add Backstage validation to example

odsod:
Newly generated repos will come with Backstage validation out of the
box.


alethenorio (commented):
@odsod Will this fail if there are no Backstage files? Where will it look for Backstage files? This is open source so should we cater for non-Einride setups somehow?

odsod:
@alethenorio By default it will look in `.backstage` and it should not fail if there are no files. Non-Einride setups can just avoid the helper!

alethenorio (approved):

Edholm (approved):

PR #444, opened 2023-07-10
feat(ghcomment): add GH TF comment with summary

loodde:
Iterated version of the TF plan PR comment functionality created by @alethenorio in the [gcp-projects repo](https://github.com/einride/gcp-projects/blob/master/.sage/cdktf.go#L361). Difference between this and the existing comment functionality is that it shows an overall summary and a list of type of resources being changed grouped by action. See example of output in image below. The full HCL plan is also included but collapsed by default. The tfplan file contains simplified TF json definitions lifted from the [github.com/hashicorp/terraform-json](github.com/hashicorp/terraform-json) repo.

image

### DIFF
--- a/tools/sgterraform/tfplan.go
+++ b/tools/sgterraform/tfplan.go
@@ -1,0 +1,126 @@
+package sgterraform
+
+// This file contains needed functionality and definitions from github.com/hashicorp/terraform-json to parse the
+// tf json plan.
+
+// Plan contains a simplified definition of the json generated when using the -json flag for terraform plan.
+// It only includes the fields needed to generate the plan summery used in CommentOnPRWithPlanSummarized.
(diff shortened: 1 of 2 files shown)


alethenorio:
@loodde I was thinking that maybe the "per change type/resource breakdown" might be less useful in some cases. A gut feeling is that in many projects (compared to gcp-projects) it would be just as simple and maybe more informative to open the plan.

What do you think about having a flag to enable/disable that part and default it as off?

Also, is the plan parser a 1/1 copy/paste from the Terraform library? If not it might be good to keep it as close as possible to it so we can bring patches/changes without having to do a lot of work every time

alethenorio:
Right. Maybe we can keep it like that and see how it develops and see if people switch over

alethenorio (approved):

PR #448, opened 2023-07-20
feat: add sgpython and sgpoetry tools

odsod:
For Python projects managed by Poetry.

### DIFF
--- a/tools/sgpython/command.go
+++ b/tools/sgpython/command.go
@@ -1,0 +1,72 @@
+package sgpython
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+
+	"go.einride.tech/sage/sg"
+	"go.einride.tech/sage/sgtool"
+	"go.einride.tech/sage/tools/sggit"
+)
+
+const (
+	name         = "python"
+	version      = "3.10.6" // parity with Ubuntu 22.04 LTS
+	pyenvVersion = "2.3.18"
+)
+
+func Command(ctx context.Context, args ...string) *exec.Cmd {
+	sg.Deps(ctx, PrepareCommand)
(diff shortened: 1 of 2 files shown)


tools/sgpython/command.go
```diff
@@ -0,0 +1,71 @@
...
+	toolDir := sg.FromToolsDir(name, version)
+	pyenvDir := filepath.Join(toolDir, "pyenv")
+	binDir := filepath.Join(pyenvDir, "versions", version, "bin")
+	pythonFromPyenv := filepath.Join(binDir, "python")
+	if _, err := os.Stat(pythonFromPyenv); err == nil {
+		if _, err := sgtool.CreateSymlink(pythonFromPyenv); err != nil {
+			return err
+		}
+		return nil
+	} else if systemPython3, err := exec.LookPath("python3"); err == nil {
```
noppaz, line 36: `python3` on system path might be a lower version than specified here. Could this get confusing if you start getting errors but expect 3.11.3 to be installed?

odsod: I was thinking on whether to check this, but since Ubuntu 22.04 LTS has Python 3.10, I think it's fairly low risk that any language features are missing that would cause issues? But we can default to 3.10.6 instead and then be on the Ubuntu LTS version!

odsod: Updated!

noppaz: Yes probably low risk of missing language features, even more so with 3.10. I think this is a good start - if the user is aware of this system Python fallback they can always check which one is used as well. It's a bit similar with the whole sage tooling depending on system Go

odsod: Good point - added a log statement that informs the user when system Python is used!

noppaz (approved):
Seems like a good starting point for Python tooling to me

PR #449, opened 2023-07-23
govulncheck helper

odsod:
- feat: add sgpython and sgpoetry tools
- feat(govulncheck): add helper for checking all modules in repo


tools/sggovulncheck/tools.go
```diff
@@ -2,27 +2,72 @@ package sggovulncheck
...
+		}
+		relativePath, err := filepath.Rel(sg.FromGitRoot(), path)
+		if err != nil {
+			return err
+		}
+		cmd := Command(
+			sg.AppendLoggerPrefix(ctx, fmt.Sprintf(" (%s): ", relativePath)),
+			"-C",
+			filepath.Dir(path),
+			"./...",
```
alethenorio, line 55: In cases where one wants to run just on the root module and specifies so in the modulePaths list, won't this cause it to run in all modules and submodules in the repo?

odsod: `./...` enumerates all packages within the current module, but a single invocation doesn't cut across mutliple modules, even if there are modules in child directories

alethenorio (approved):

PR #450, opened 2023-07-24
feat: add trivy tool

alethenorio:
tfsec is [deprecated](https://github.com/aquasecurity/tfsec/discussions/1994) and the recommendation is to use trivy where the development will continue.
Trivy does a lot more than just Terraform security checking but this commit is limited to replacing tfsec and includes only one helper command to make the transition easy for `sgtfsec` users (API is slightly different).

`sgtrivy` also includes a default .trivyignore which currently only ignores https://avd.aquasec.com/misconfig/google/storage/avd-gcp-0066/ for being too strict.

Lastly, the sgtfsec package has been marked as deprecated.

### DIFF
--- a/tools/sgtrivy/tools.go
+++ b/tools/sgtrivy/tools.go
@@ -1,0 +1,99 @@
+package sgtrivy
+
+import (
+	"context"
+	_ "embed"
+	"errors"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+
+	"go.einride.tech/sage/sg"
+	"go.einride.tech/sage/sgtool"
+)
+
+//go:embed trivyignore
+var defaultConfig []byte
+
+const (
+	version = "0.43.1"
+	name    = "trivy"
+)
+
+func defaultConfigPath() string {
+	return sg.FromToolsDir(name, ".trivyignore")
+}
+
+// CheckTerraformCommand checks terraform configuration on the given dir
+// for any known security misconfigurations.
+// It includes a default .trivyignore.yaml which can be
+// overridedn by setting a .trivyignore.yaml in the git root.
+func CheckTerraformCommand(ctx context.Context, dir string) *exec.Cmd {
+	args := []string{
+		"config",
+		"--exit-code",
+		"1",
+		dir,
+	}
+
+	configPath := sg.FromGitRoot(".trivyignore")
+	if _, err := os.Lstat(configPath); errors.Is(err, os.ErrNotExist) {
+		configPath = defaultConfigPath()
+	}
+	args = append(args, "--ignorefile", configPath)
+
+	return Command(ctx, args...)
+}
+
+func Command(ctx context.Context, args ...string) *exec.Cmd {
+	sg.Deps(ctx, PrepareCommand)
+	return sg.Command(ctx, sg.FromBinDir(name), args...)
+}
+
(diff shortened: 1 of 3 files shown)


tools/sgtrivy/tools.go
```diff
@@ -0,0 +1,99 @@
...
+	name    = "trivy"
+)
+
+func defaultConfigPath() string {
+	return sg.FromToolsDir(name, ".trivyignore")
+}
+
+// CheckTerraformCommand checks terraform configuration on the given dir
+// for any known security misconfigurations.
+func CheckTerraformCommand(ctx context.Context, dir string) *exec.Cmd {
```
odsod, line 31: IMO since this helper only adds a few arguments, might be better to inline this at the call sites to make it more clear how the command is invoked?

tools/sgtrivy/tools.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	return Command(ctx, args...)
+}
+
+// Command returns a `trivy` *exec.Cmd.
+// It includes a flag to use a default .trivyignore.yaml which can be
+// overridedn by setting a .trivyignore.yaml in the git root.
+func Command(ctx context.Context, args ...string) *exec.Cmd {
+	sg.Deps(ctx, PrepareCommand)
```
odsod, line 46: The `Command` function should always return a pure command without any alterations - but if we move the `.trivyignore` stuff up to the helper above, then the helper is a bit more justified!

tools/sgtrivy/tools.go
```diff
@@ -0,0 +1,99 @@
...
+	"os/exec"
+	"path/filepath"
+	"runtime"
+
+	"go.einride.tech/sage/sg"
+	"go.einride.tech/sage/sgtool"
+)
+
+//go:embed trivyignore
+var DefaultConfig []byte
```
odsod, line 18: IMO hide this until someone really needs it, in case we want to change the public API?

tools/sgtrivy/tools.go
alethenorio: Sure. It was mostly to make the transition from tfsec easier

odsod (approved):
LGTM, but suggesting to move the `--ignorefile` arguments up to the helper function and keep the `Command` function to return a pure command

tools/sgtrivy/tools.go
alethenorio: Good point. Copy/Paste from another package

tools/sgtrivy/tools.go
alethenorio: Make sense but the `.trivyignore` file isn't necessarily specific to Terraform so it wouldn't quite make sense to put it in the helper. I tried following the same pattern we have in other packages though I see now that mostly we do this in the `Run()` APIs which we don't have here.

If we remove it from here then I don't think it makes sense to supply the file if the caller has to write the logic to use.

In such case I think it is better to keep the Terraform helper and wire it there. I'll change the PR

tools/sgtrivy/tools.go
alethenorio: I've moved the .tryviignore stuff in here now so keeping it

odsod: Nice!

tools/sgtrivy/tools.go
odsod: Sounds good!

PR #451, opened 2023-07-28
fix(sgtrivy): go arch for arm64

backevik:


alethenorio (approved):
Doh. Missed this. Thanks

PR #453, opened 2023-08-04
fix: init sage in repository not working

antonmoller:
If merged this PR will fix the init command returning the error `.sage: file exists`. This was caused by a recent addition of `ensureParentDir` to `FromBinDir`. The function `sg.Command` uses `FromBinDir` which causes the .sage directory to already be created when `os.Mkdir` is called in `initSage`
image

### DIFF
--- a/main.go
+++ b/main.go
@@ -137,7 +137,8 @@ func resolveSageModulePath(ctx context.Context) (string, error) {
 		return moduleName, nil //nolint:nilerr
 	}
 	var out bytes.Buffer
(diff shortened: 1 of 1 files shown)


odsod (approved):
Nice catch! Maybe we should also fix the `ensureParentDirectory` function to not create the parent dir if it already exists?

antonmoller:
> Nice catch! Maybe we should also fix the `ensureParentDirectory` function to not create the parent dir if it already exists?

Sounds like a good idea I can look into it

PR #458, opened 2023-08-18
Add 'build' to commit-lint ignore commit tag

joaoeinride:
Adds 'build' to ignored commit tag for dependabot.

Reason to add is https://github.com/einride/map-go/pull/343

Let me know if there is a better way of doing this :)


Edholm:
It is is probably better if you configure the [`commit-message`](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#commit-message) config for dependabot in that repo.

e.g. 
```
    commit-message:
      prefix: chore
      include: scope
```

joaoeinride:
that should work. closing this as unecessary

PR #465, opened 2023-09-29
feat: bump api-linter to 1.57.1

kallekrantz:
Somehow 1.52.4 version causes a segfault in some localized cases. Cannot replicate the segfault on 1.57.1.


Edholm (approved):
I was actually thinking about this just this morning (but for different reasons).

PR #466, opened 2023-10-03
fix: rollback version update on api-linter

DavyJ0nes:
Updates api-linter to version [v1.56.1](https://github.com/googleapis/api-linter/releases/tag/v1.56.1) as v.[1.57.0](https://github.com/googleapis/api-linter/releases/tag/v1.57.0) introduced a new
requirement for (google.api.field_behavior) = IDENTIFIER to be set on
the name field of a proto resource.

There is a conflict between buf lint and api-linter due to an update to
field_behaviour.proto. An issue has been opened buf as fix is on their side:
https://github.com/bufbuild/buf/issues/2474

PR #467, opened 2023-10-03
feat: update api-linter version to v1.57.1

DavyJ0nes:
The issue referenced in #466 has been resolved so we can update the api-linter version again.


Edholm (approved):
Nice!

PR #468, opened 2023-10-12
feat: add action input of cache-dependency-path

noppaz:
By enabling changing the go.sum file(s) to cache on users can configure which subdir is important for their respective actions.

The main use case would most likely be to cache the dependencies of sage in repositories that have many dependencies for sage itself, but it could also be used to cache dependencies in multi module repositories. This is especially useful for sage targets that doesn't run go code but instead a downloaded binary (e.g., Terraform). Some tests of internal repositories halves the time of execution as sage dependency installation is a big part of the runtime of the action.

The default is set to what [setup-go has as default](https://github.com/actions/setup-go#caching-dependency-files-and-build-outputs); the root go.sum in the repository.

### DIFF
--- a/actions/setup/action.yml
+++ b/actions/setup/action.yml
@@ -17,6 +17,11 @@ inputs:
     required: false
     default: '1.21'
 
+  cache-dependency-path:
+    description: Path to the go.sum file(s) to use in caching.
+    required: false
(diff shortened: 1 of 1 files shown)


Edholm (approved):
Neat :+1:

PR #470, opened 2023-10-24
feat(sgdocker): support apple silicon

josh-richmond:
Previous when running from a mac with apple silicon you will get routed too:
`https://download.docker.com/mac/static/stable/arm64/...`

With this change it will correctly route too:
`https://download.docker.com/mac/static/stable/aarch64/...`


Edholm (approved):
Nicely spotted :)

PR #473, opened 2023-11-01
feat(cloud-endpoints): bump ESPv2 to 2.46.0

philiphassel:
Go version updated to 1.20 which should fix CVE-2022-27664:

>  In net/http in Go before 1.18.6 and 1.19.x before 1.19.1, attackers can cause a denial of service because an HTTP/2 connection can hang during closing if shutdown were preempted by a fatal error.


Edholm (approved):
Nice!

PR #475, opened 2023-11-06
feat(sgcommitlint): bump commitlint to v18

filiptammergard:
Dropping support for Node v16 and TypeScript v4.


Edholm (approved):
Any breaking changes we need to be aware of? :)

filiptammergard:
@Edholm Only dropping support for Node v16 and TypeScript v4, so if you're decently up to date dependency-wise, there should be no problems upgrading 🙂

PR #477, opened 2023-11-09
feat: add tool for authenticating npm to artifact registry

filiptammergard:


tools/sgnpmartifactregistryauth/tools.go
```diff
@@ -0,0 +1,26 @@
...
+	sg.Deps(ctx, PrepareCommand)
+	return sg.Command(ctx, sg.FromBinDir(name), args...)
+}
+
+func PrepareCommand(ctx context.Context) error {
+	sg.Logger(ctx).Println("authenticating npm to artifact registry...")
+	return sg.Command(
+		ctx,
+		"npx",
+		"google-artifactregistry-auth@3.1.2",
```
alethenorio, line 24: Maybe break the version into a constant and do something like

```suggestion
		fmt.Sprintf("google-artifactregistry-auth@%s", version),
```

alethenorio (approved):
Nice. Small comment but looks good

tools/sgnpmartifactregistryauth/tools.go
filiptammergard: Sounds good!

PR #486, opened 2023-11-15
feat: switch markdown formatter

DavyJ0nes:
This changes the default markdown formatter from `markdownfmt` to `mdformat`. The reason to do this is for numerous reasons, the main ones are:
- Better support for GH flavour of markdown, which most users are used to and there are less surprises when formatted and rendered
- Supports reference style links, see issue: https://github.com/shurcooL/markdownfmt/issues/34
- Has support for line wrapping to make md files easier to read in raw format in terminal.

I have also added default arguments to mdformat to make it easier to adopt without having to specify these args yourself but can overwrite them as needed.

### DIFF
--- a/.sage/main.go
+++ b/.sage/main.go
@@ -11,7 +11,7 @@ import (
 	"go.einride.tech/sage/tools/sggolangcilint"
 	"go.einride.tech/sage/tools/sggolicenses"
 	"go.einride.tech/sage/tools/sggoreview"
-	"go.einride.tech/sage/tools/sgmarkdownfmt"
+	"go.einride.tech/sage/tools/sgmdformat"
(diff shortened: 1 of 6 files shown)


odsod (approved):
Super good!!

viktorvoltaire:
Jay

Edholm:
Switching to this formatter in our repo makes formatting files take roughly a minute on my machine and uses 100% cpu. It looks like recursively searching for markdown files is very inefficient in large repos :crying_cat_face:

DavyJ0nes:
I've updated the default args so that you can now specify all the args to mdformat and not always run it recursively over all files. That should help make it a bit easier to use.

PR #488, opened 2023-11-17
feat: add new sgartifactregistry package

alethenorio:
Include NPMAuthenticate target to authenticate towards npm artifact
registries.

We authenticate using both npm and yarn

relates to [TCC-345](https://einride.atlassian.net/browse/TCC-345)

[TCC-345]: https://einride.atlassian.net/browse/TCC-345?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tools/sgartifactregistry/auth.go
+++ b/tools/sgartifactregistry/auth.go
@@ -1,0 +1,81 @@
+package sgartifactregistry
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"os/exec"
+	"strings"
+
+	"go.einride.tech/sage/sg"
+	"go.einride.tech/sage/tools/sggcloud"
+)
+
+// NpmAuthenticate authenticates to the Artifact Registry in the give registryURL.
+// If yarn major version is 1 we add authentication information for the given registryURL
+// in .npmrc in the user directory otherwise add to .yarnrc.yaml in the user directory.
+// The npm and yarn commands are run from the give package.json directory in order to
+// use the proper yarn version.
+func NpmAuthenticate(ctx context.Context, packageJSONDir, registryURL string) error {
+	// Grab Google Cloud Access Token
+	var accessTokenOutput strings.Builder
+	cmd := sggcloud.Command(ctx, "auth", "print-access-token")
+	cmd.Stdout = &accessTokenOutput
+	if err := cmd.Run(); err != nil {
+		return err
+	}
+	registry := strings.TrimPrefix(registryURL, "https://")
+
+	// If we have yarn installed, find its version
+	yarnMajor := "1"
+	_, err := exec.LookPath("yarn")
+	if err != nil {
+		if !errors.Is(err, exec.ErrNotFound) {
+			return err
+		}
+	} else {
+		// Find yarn version
+		cmd = sg.Command(
+			ctx,
+			"yarn",
+			"--version",
+		)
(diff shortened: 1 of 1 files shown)


tools/sgartifactregistry/auth.go
```diff
@@ -0,0 +1,53 @@
...
+	"context"
+	"fmt"
+	"io"
+	"strings"
+
+	"go.einride.tech/sage/sg"
+	"go.einride.tech/sage/tools/sggcloud"
+)
+
+// Develop starts the Cloud Run service at the provided Go path with the provided service account and config.
```
filiptammergard, line 13: Is this comment a copy-paste mistake?

tools/sgartifactregistry/auth.go
```diff
@@ -0,0 +1,53 @@
...
+func NPMAuthenticate(ctx context.Context, registryURL string) error {
```
filiptammergard, line 14: Doesn't matter much, but looks slightly better IMO 🤠 

```suggestion
func NpmAuthenticate(ctx context.Context, registryUrl string) error {
```

tools/sgartifactregistry/auth.go
```diff
@@ -0,0 +1,53 @@
...
+		return err
+	}
+	// Set yarn config
+	cmd = sg.Command(
+		ctx,
+		"yarn",
+		"config",
+		"set",
+		"--home",
+		"npmRegistries[\"//europe-west1-npm.pkg.dev/e-npm-registry-prod/npm\"].npmAuthToken",
```
filiptammergard, line 46: Shouldn't this also use `registry`?

tools/sgartifactregistry/auth.go
alethenorio: Indeed. Fixed :)

tools/sgartifactregistry/auth.go
alethenorio: Go is a little specific on abbreviations so URL is more common (and the linter will scream about it otherwise). I can change the npm one.

tools/sgartifactregistry/auth.go
alethenorio: Oops that was a copy/paste error. Fixed it

tools/sgartifactregistry/auth.go
filiptammergard: Ah I see, go ahead with what's more common 👍

tools/sgartifactregistry/auth.go
```diff
@@ -0,0 +1,68 @@
...
+	cmd = sg.Command(
+		ctx,
+		"npm",
+		"config",
+		"set",
+		"-L",
+		"user",
+		fmt.Sprintf("//%s/:_authToken", registry),
+		strings.TrimSpace(accessTokenOutput.String()),
+	)
```
filiptammergard, line 36: I'm thinking this should do the trick for both NPM and Yarn 1 users

alethenorio: Good comment. Changed to a switch statement instead. WDYT?

filiptammergard (approved):
🙏

thall (approved):

PR #489, opened 2023-11-20
fix: allow paths arg to be updated on mdformat

DavyJ0nes:
Previously it was not possible to specify the paths that mdformat should be run against due to the way that the default args were set. This allows either all args to be replaced or use the default thus giving users more control.


Edholm (approved):
Noice!

alethenorio (approved):

PR #492, opened 2023-11-22
feat(sgartifactregistry): determine if access token is expired on npm auth

alethenorio:
When using `gcloud auth print-access-token`, gcloud will print the token
regardless on whether the token is expired or not.
We use gcloud to figure out if the token is expired for a better user
experience (instead of the user getting back a 403 which can be
confusing)

### DIFF
--- a/tools/sgartifactregistry/auth.go
+++ b/tools/sgartifactregistry/auth.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"os/exec"
+	"strconv"
 	"strings"
 
 	"go.einride.tech/sage/sg"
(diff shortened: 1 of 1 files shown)


Edholm (approved):
:+1:

filiptammergard (approved):

tools/sgartifactregistry/auth.go
```diff
@@ -24,13 +25,22 @@ func NpmAuthenticate(ctx context.Context, packageJSONDir, registryURL string) er
...
 	registry := strings.TrimPrefix(registryURL, "https://")
 	// Trailing slashes at the end of the URL have been known to cause issues with some setups
 	registry = strings.TrimSuffix(registry, "/")
 
+	expired, err := isGoogleAuthExpired(ctx)
+	if err != nil {
+		return fmt.Errorf("unable to verify Google authentication token expiration: %v", err)
+	}
+	if expired {
+		return fmt.Errorf("google authentication token is expired. Please authenticate using 'gcloud'")
```
filiptammergard, line 38: Should we add what exact commands to use to authenticate? Just to make it a little bit easier for someone that doesn't know the recommended way of authenticating?

alethenorio: Yes. Can fix

PR #493, opened 2023-11-22
feat(mdformat): only consider markdown files known to Git

Edholm:
The recursive argument to mdformat (`.`) is quite slow in repos with lots of files and folders where no markdown files are expected to be found. We now instead search for all markdown files that has been checked into the repo (plus untracked files), instead of letting `mdformat` do that.

If you want the previous behaviour you have to pass your own arguments to `Command`.

### DIFF
--- a/tools/sgmdformat/tools.go
+++ b/tools/sgmdformat/tools.go
@@ -5,9 +5,11 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strings"
 
 	"go.einride.tech/sage/sg"
 	"go.einride.tech/sage/sgtool"
+	"go.einride.tech/sage/tools/sggit"
 	"go.einride.tech/sage/tools/sgpython"
 )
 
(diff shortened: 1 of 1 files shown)


Edholm:
This solution is roughly 60s faster in one of our repos :)

odsod:
Can we package this in an even simpler way so that users still just have to run a single task that takes care of all this with good defaults?

Edholm:
> Can we package this in an even simpler way so that users still just have to run a single task that takes care of all this with good defaults?

Makes sense. Now that I think about it we might as well replace the `.` and find markdown files ourselves instead by default.

Edholm:
@alethenorio  @odsod  have a look now. Based the solution on `ls-files` like suggested and made that behaviour the default.

Much neater now if I do say so myself

tools/sgmdformat/tools.go
```diff
@@ -60,3 +63,26 @@ func PrepareCommand(ctx context.Context) error {
...
+// list markdown files known by git + untracked ones.
+func listMarkdownFiles(ctx context.Context) []string {
+	output := strings.TrimSpace(
+		sg.Output(sggit.Command(
+			ctx,
+			"ls-files",
+			"*.md",               // only markdown files
+			"--others",           // include untracked files
+			"--exclude-standard", // exclude ignored files
+			"--cached",           // include "normal" files
```
alethenorio, line 76: So this excludes any .gitignored files and allows everything else?

Is `--cached` needed?

tools/sgmdformat/tools.go
```diff
@@ -60,3 +63,26 @@ func PrepareCommand(ctx context.Context) error {
...
+		)),
+	)
+	if len(output) == 0 {
+		return nil
+	}
+	files := strings.Split(output, "\n")
+	for i, file := range files {
+		// Convert to absolute path
+		files[i] = sg.FromGitRoot(file)
```
alethenorio, line 85: Will this call git for every single file? Could this be a performance concern?

alethenorio (approved):
Super nice. Got a few comments but nothing major so approving it.

tools/sgmdformat/tools.go
Edholm: Yes, cached is the default flag if you don't invoke e.g. --others. So since we use --others, then we also need to include --cached to get all files, not just untracked ones.

tools/sgmdformat/tools.go
Edholm: On my machine with 5 md files to format, this step takes 4-5ms so quite fast. That said though it is not needed and was more of a "hängslen and livrem" kinda thing. 

I will remove it :)

alethenorio: It was a great idea. We can introduce it again if we run into issues.

PR #494, opened 2023-11-27
mdformat: setup dependency file and add admonitions support

radhus:
This tracks Python dependencies in a `requirements.txt` file to easier handle multiple dependencies (likely only mdformat plugins).

Also, add the [admonitions mdformat plugin](https://github.com/KyleKing/mdformat-admon) to support [TechDocs admonitions](https://squidfunk.github.io/mkdocs-material/reference/admonitions/#admonitions).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -20,3 +20,8 @@ updates:
     directory: .sage
     schedule:
       interval: weekly
+
+  - package-ecosystem: pip
+    directory: tools/sgmdformat
+    schedule:
+      interval: weekly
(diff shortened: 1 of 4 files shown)


tools/sgmdformat/tools_test.go
```diff
@@ -0,0 +1,13 @@
+package sgmdformat
+
+import (
+	"context"
+	"testing"
+)
+
+func TestDepedencies(t *testing.T) {
+	err := PrepareCommand(context.TODO())
```
alethenorio, line 9: Maybe use context.Background instead?

alethenorio (approved):

PR #498, opened 2023-12-18
feat: add Backstage Go coverage reporting (and gcov2lcov tool)

maxekman:
Add Backstage coverage reporting command and gcov2lcov tool needed for it.

https://github.com/jandelgado/gcov2lcov

### DIFF
--- a/tools/sgbackstagecoverage/tools.go
+++ b/tools/sgbackstagecoverage/tools.go
@@ -1,0 +1,38 @@
+package sgbackstagecoverage
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"os"
+)
+
+// Post posts code coverage in LCOV format to the Backstage API.
+// apiURL is the URL to the API, for example 'https://api.mybackstagedeployment.com'.
+// authToken is the Bearer token for the Backstage API.
+// coverFile is the coverage file in LCOV format (use sggcov2lcov to convert Go test output).
+// entity is the entity ref, for example 'component:default/my-service'.
+func Post(ctx context.Context, apiURL, authToken, coverFile, entity string) error {
+	url := fmt.Sprintf("%s/api/code-coverage/report?entity=%s&coverageType=lcov", apiURL, entity)
+	f, err := os.Open(coverFile)
+	if err != nil {
+		return fmt.Errorf("opening coverage file: %w", err)
+	}
+	defer f.Close()
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, url, f)
+	if err != nil {
+		return fmt.Errorf("creating coverage POST request: %w", err)
+	}
+	req.Header.Set("Content-Type", "text/plain")
+	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", authToken))
+	client := &http.Client{}
+	resp, err := client.Do(req)
+	if err != nil {
+		return fmt.Errorf("posting coverage data: %w", err)
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusCreated {
+		return fmt.Errorf("coverage post status not OK: %s", http.StatusText(resp.StatusCode))
+	}
+	return nil
+}
--- a/tools/sggcov2lcov/tools.go
+++ b/tools/sggcov2lcov/tools.go
@@ -1,0 +1,64 @@
+package sggcov2lcov
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"strings"
+
+	"go.einride.tech/sage/sg"
+	"go.einride.tech/sage/sgtool"
+)
+
+const (
+	name    = "gcov2lcov"
+	version = "1.0.6"
+)
+
+func Command(ctx context.Context, args ...string) *exec.Cmd {
+	sg.Deps(ctx, PrepareCommand)
+	cmd := sg.Command(ctx, sg.FromBinDir(name), args...)
+	// GOROOT need to be set, and as suggested here: https://github.com/jandelgado/gcov2lcov#goroot.
+	goroot := sg.Output(sg.Command(ctx, "go", "env", "GOROOT"))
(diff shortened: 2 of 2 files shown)


tools/sgbackstagecoverage/tools.go
```diff
@@ -0,0 +1,39 @@
...
+
+	"go.einride.tech/sage/sg"
+	"go.einride.tech/sage/tools/sggcov2lcov"
+)
+
+func PostCoverage(ctx context.Context, entity string) error {
+	if err := sggcov2lcov.Command(ctx).Run(); err != nil {
+		return fmt.Errorf("converting coverage format from gcov to lcov: %w", err)
+	}
+	url := fmt.Sprintf("https://api.roadie.so/api/code-coverage/report?entity=%s&coverageType=lcov", entity)
```
maxekman, line 17: Should the Backstage URL be a parameter here?

maxekman: Made it a parameter, together with the auth env.

tools/sggcov2lcov/tools.go
```diff
@@ -0,0 +1,70 @@
...
+
+const (
+	name    = "gcov2lcov"
+	version = "1.0.6"
+)
+
+func Command(ctx context.Context, args ...string) *exec.Cmd {
+	sg.Deps(ctx, PrepareCommand)
+	coverFile := sg.FromBuildDir("go", "coverage", "go-test.txt")
+	lcovFile := sg.FromBuildDir("go", "coverage", "go-test.lcov")
```
alethenorio, line 24: The convention in Sage is generally to keep the command as clean as possible running nothing but the tool and leaving any conventions with the tool to more specific commands. Can we leave these out of this command?

tools/sggcov2lcov/tools.go
```diff
@@ -0,0 +1,70 @@
...
+		append(
+			[]string{
+				"-infile=" + coverFile,
+				"-outfile=" + lcovFile,
+			},
+			args...,
+		)...,
+	)
+	goroot := sg.Output(sg.Command(ctx, "go", "env", "GOROOT"))
+	cmd.Env = append(cmd.Env, "GOROOT="+goroot)
```
alethenorio, line 37: Why is this needed? Setting the GOROOT is something one should generally not need

tools/sgbackstagecoverage/tools.go
```diff
@@ -0,0 +1,44 @@
...
+
+	"go.einride.tech/sage/sg"
+	"go.einride.tech/sage/tools/sggcov2lcov"
+)
+
+// PostGoCoverage posts Go code coverage to the Backstage API.
+// apiURL is the URL to the API, for example 'https://api.mybackstagedeployment.com'.
+// entity is the entity ref, for example 'component:default/my-service'.
+// authEnv is the env var with the Bearer, for example 'MY_BACKSTAGE_TOKEN'.
+func PostGoCoverage(ctx context.Context, apiURL, entity, authEnv string) error {
```
alethenorio, line 17: What do you think about passing the token directly and letting the caller decide whether that lives in an environment variable or a file or something else?

tools/sggcov2lcov/tools.go
maxekman: Good point, I have moved these to be arguments instead.

tools/sggcov2lcov/tools.go
maxekman: It was needed when I ran the command. Not sure about the details, but it was mentioned in the tool docks so there is probably a valid reason: https://github.com/jandelgado/gcov2lcov#goroot

tools/sgbackstagecoverage/tools.go
maxekman: Moved it to an argument instead, and also removed the dependency on the `sggcov2lcov` command making this a pure LCOV post command.

tools/sggcov2lcov/tools.go
alethenorio: Did we try without it? The docs mention it might be needed and it would be great to skip it as it could lead to issues in some setups.

maxekman: Yes, otherwise I would not have added it.

maxekman: Do you have an idea how we can make it optional instead if you prefer?

tools/sggcov2lcov/tools.go
alethenorio: The command  target should be a clean target and run nothing else but the command and given args. No assumptions about any flags should be made in here.

Either we should use the infile and outfile arguments locally in each project or break it out to a separate target

tools/sgbackstagecoverage/tools.go
```diff
@@ -0,0 +1,38 @@
...
+	"net/http"
+	"os"
+)
+
+// PostCoverage posts code coverage in LCOV format to the Backstage API.
+// apiURL is the URL to the API, for example 'https://api.mybackstagedeployment.com'.
+// authToken is the Bearer token for the Backstage API.
+// coverFile is the coverage file in LCOV format (use sggcov2lcov to convert Go test output).
+// entity is the entity ref, for example 'component:default/my-service'.
+func PostCoverage(ctx context.Context, apiURL, authToken, coverFile, entity string) error {
```
alethenorio, line 15: We should keep in mind Go readability conventions here. With this target name and package name we'll get `sgbackstagecoverage.PostCoverage`. Maybe rename the target to just `Post`?

maxekman: Sure, good idea.

tools/sggcov2lcov/tools.go
alethenorio: You can let the user simply set the environment variable on the local project. Will the tool even work without it though? I am mostly trying to avoid future issues. Having to set GOROOT feels a little like working against the Go ecosystem but if the tool requires it to work I don't see any harm in keeping it for now. Maybe add a comment on it?

alethenorio (approved):

PR #516, opened 2024-02-25
fix(sg): remove -fm suffix in target names

ericwenn:
It seems common to pass arguments to Deps and SerialDeps as below:
```go
type namespace sg.Namespace

func (n namespace) Default(ctx context.Context) error {
            sg.Deps(ctx, n.FirstTarget, n.SecondTarget)
                    return nil
}

```

Currently this leads to logs in the dep targets to be prefixed with
`[namespace:first-target-fm]` and `[namespace:second-target-fm]`, because
of how the compiler implements methods bound to specific receivers. If
instead the dependencies were specified as:
```go
sg.Deps(ctx, namespace.FirstTarget, namespace.SecondTarget)
```
the names would log prefixes would simply be `[namespace:first-target]`
and `[namespace:second-target]`.

This commit removes the `-fm` suffix. This is safe as no function name
in go can include a hyphen.

### DIFF
--- a/sg/fn.go
+++ b/sg/fn.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"reflect"
 	"runtime"
+	"strings"
 )
 
 // Target represents a target function that can be run with Deps.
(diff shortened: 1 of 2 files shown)


thall (approved):
Nice! 👌🚀

odsod (approved):
Nice find! :star:

PR #519, opened 2024-02-28
Add golines

fredrikaverpil:
https://github.com/segmentio/golines

### DIFF
--- a/tools/sggolines/tools.go
+++ b/tools/sggolines/tools.go
@@ -1,0 +1,57 @@
+package golines
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+
+	"go.einride.tech/sage/sg"
+	"go.einride.tech/sage/sgtool"
+)
+
+const (
+	name    = "golines"
+	version = "0.12.2"
+)
+
+// Command returns an [*exec.Cmd] for golines.
+func Command(ctx context.Context, args ...string) *exec.Cmd {
+	sg.Deps(ctx, PrepareCommand)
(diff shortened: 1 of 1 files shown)


tools/sggolines/tools.go
```diff
@@ -0,0 +1,28 @@
...
+)
+
+// Command returns an [*exec.Cmd] for golines.
+func Command(ctx context.Context, args ...string) *exec.Cmd {
+	sg.Deps(ctx, PrepareCommand)
+	return sg.Command(ctx, sg.FromBinDir(name), args...)
+}
+
+func PrepareCommand(ctx context.Context) error {
+	_, err := sgtool.GoInstall(ctx, "github.com/segmentio/golines", version)
```
odsod, line 23: IMO since golines publishes pre-built artifacts let's use those:
https://github.com/segmentio/golines/releases

odsod: (faster to install, especially on GitHub Actions when it's fetched from within the same server rack :smile:)

odsod (commented):
Great initiative!

IMO we could add `GoFormat` as a default target to the Sage template, but then we should also run GolangCI-Lint in fix mode (which should fix imports, protogetters, and whatnot)

tools/sggolines/tools.go
fredrikaverpil: Fixed... but I believe github is having issues updating the PR: https://githubstatus.com

fredrikaverpil: Ok, now it looks like the PR got updated. 😆

odsod (approved):

PR #521, opened 2024-03-01
Add gci

fredrikaverpil:


tools/sggci/tools.go
```diff
@@ -0,0 +1,28 @@
...
+)
+
+// Command returns an [*exec.Cmd] for golines.
+func Command(ctx context.Context, args ...string) *exec.Cmd {
+	sg.Deps(ctx, PrepareCommand)
+	return sg.Command(ctx, sg.FromBinDir(name), args...)
+}
+
+func PrepareCommand(ctx context.Context) error {
+	_, err := sgtool.GoInstall(ctx, "github.com/daixiang0/gci", version)
```
odsod, line 23: this is distributed with GolangCI-Lint - do we need a separate installation of it?

PR #524, opened 2024-03-05
Add option to not read golangci-lint config yaml

fredrikaverpil:
### Why?

I would like to be able to use a selection of linters only.

### What?

Add ability to read `--no-config` (this is an actual golangci-lint commandline option, but cannot be combined with `--config`).

### Notes

The only caveat I can come to think of is we have to define certain configs in both yaml and in Go, such as `gofumpt`'s `extra-rules`.


tools/sggolangcilint/tools.go
```diff
@@ -33,15 +33,32 @@ func defaultConfigPath() string {
 }
 
 func CommandInDirectory(ctx context.Context, directory string, args ...string) *exec.Cmd {
-	configPath := filepath.Join(directory, ".golangci.yml")
-	if _, err := os.Lstat(configPath); errors.Is(err, os.ErrNotExist) {
-		configPath = defaultConfigPath()
+	skipConfig := false
```
odsod, line 36: var skipConfig bool

https://dave.cheney.net/practical-go/presentations/qcon-china.html#_use_a_consistent_declaration_style

tools/sggolangcilint/tools.go
```diff
@@ -33,15 +33,32 @@ func defaultConfigPath() string {
...
 
 func CommandInDirectory(ctx context.Context, directory string, args ...string) *exec.Cmd {
-	configPath := filepath.Join(directory, ".golangci.yml")
-	if _, err := os.Lstat(configPath); errors.Is(err, os.ErrNotExist) {
-		configPath = defaultConfigPath()
+	cmdArgs := []string{"run"}
+
+	skipConfig := false
+	for _, arg := range args {
+		if arg == "--no-config" {
```
odsod, line 40: not important to do now, but let's bump to Go 1.22 and use slices.Contains? :smile_cat:

odsod: then you might even inline the whole variable and just say `if slices.Contains(...)`

tools/sggolangcilint/tools.go
fredrikaverpil: I updated the code, which is now using `:=` consistently.

tools/sggolangcilint/tools.go
fredrikaverpil: We have panics with 1.22  in book-service, so that requires investigations. I would rather do that separately.

tools/sggolangcilint/tools.go
fredrikaverpil: Ok, I thought you meant to use some consistent declaration style... 😄 
I made the skipConfig into a var.

tools/sggolangcilint/tools.go
```diff
@@ -33,15 +33,31 @@ func defaultConfigPath() string {
...
+		cmdArgs = append(cmdArgs, "--no-config")
 	}
+
 	var excludeArg []string
 	if directory == sg.FromSageDir() {
-		excludeArg = append(excludeArg, "--exclude", "(is a global variable|is unused)")
+		cmdArgs = append(cmdArgs, "--exclude", "(is a global variable|is unused)")
 	}
-	cmdArgs := append([]string{"run", "-c", configPath}, args...)
+
```
odsod, line 59: same comment as in a previous review: omit adding blank lines based on taste

fredrikaverpil: Fixed.

tools/sggolangcilint/tools.go
```diff
@@ -33,16 +33,27 @@ func defaultConfigPath() string {
...
+		cmdArgs = append(cmdArgs, "--no-config")
 	}
-	var excludeArg []string
 	if directory == sg.FromSageDir() {
-		excludeArg = append(excludeArg, "--exclude", "(is a global variable|is unused)")
+		cmdArgs = append(cmdArgs, "--exclude", "(is a global variable|is unused)")
 	}
-	cmdArgs := append([]string{"run", "-c", configPath}, args...)
-	cmd := Command(ctx, append(cmdArgs, excludeArg...)...)
+	cmd := Command(ctx, append(cmdArgs, args...)...)
```
fredrikaverpil, line 56: @odsod notice how I changed the way that `--exclude` was earlier placed _after_ the user-supplied arguments (`args`), but I'm changing that here. Do you think that's a problem?

```diff
- cmd := Command(ctx, append(cmdArgs, excludeArg...)...)
+ cmd := Command(ctx, append(cmdArgs, args...)...)
```

odsod: Shouldn't be a problem, but let's test in your services first?

alethenorio:
@fredrikaverpil What about overriding the default config with your own? That should allow you to achieve the same result no?

fredrikaverpil:
> @fredrikaverpil What about overriding the default config with your own? That should allow you to achieve the same result no?

Yeah, something like that is likely required/desired, because I cannot specify linter options on the commandline (see [here](https://github.com/einride/book-service/pull/1681#discussion_r1512773140)).

PR #525, opened 2024-03-05
Support custom golangci-lint yaml config

fredrikaverpil:


alethenorio (commented):
Is this so you can specify a config file in a different place than the root of the directory supplied in the argument?

fredrikaverpil:
> Is this so you can specify a config file in a different place than the root of the directory supplied in the argument?

I'm right now calling golangci-lint like this, from another repo (which also hosts the config file at the moment):

```go
sggolangcilint.Run(ctx, "--config", sg.FromGitRoot(".sage", "golangci_format.yml"), "--fix")
```

This might not be ideal if we want to make use of this in other repos. Then we might want to store this variant of the config in the sage repo. I'm putting this PR into draft.

fredrikaverpil:
I'm keeping this "private" during our evaluation of running golines on filesave and in CI. We can discuss if/how we want to move some of this logic/config to Sage.

PR #526, opened 2024-03-06
feat(golangci-lint): add Fix helper

odsod:
Runs `golangci-lint --fix` with a default fix config which is a subset
of the default lint config, with a few modifications:

* Don't run in parallell (may cause issues between fixers)
* Remove all linters who are not fixers (to speed up invocation)

### DIFF
--- a/tools/sggolangcilint/tools.go
+++ b/tools/sggolangcilint/tools.go
@@ -23,6 +23,9 @@ const (
 //go:embed golangci.yml
 var DefaultConfig []byte
 
+//go:embed golangci.fix.yml
+var defaultFixConfig []byte
+
 func Command(ctx context.Context, args ...string) *exec.Cmd {
 	sg.Deps(ctx, PrepareCommand)
 	return sg.Command(ctx, sg.FromBinDir(name), args...)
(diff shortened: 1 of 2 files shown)


fredrikaverpil (approved):
Nice - this looks great to me. 🚀 
Let's try it?

tools/sggolangcilint/tools.go
```diff
@@ -32,6 +35,10 @@ func defaultConfigPath() string {
 	return sg.FromToolsDir(name, ".golangci.yml")
 }
 
+func defaultFixConfigPath() string {
+	return sg.FromToolsDir(name, ".golangci.fix.yml")
```
fredrikaverpil, line 39: In the future, perhaps we want two separate yaml files; one that runs linters with comes with auto-fix capabilities (modifies files in-place), and one that just runs checks.

Nit: maybe refer to this file as `.golangci.auto-fix.yml`?

fredrikaverpil: Actually, scratch this on isolating linters into _either_ `.golangci.fix.yml` or `.golangci.check.yml` files.

In the editor (Neovim), I want to be able to point my linter to look at ONE file (I don't think I can tell it to use two files) and then I want _all_ linters in that file.

PR #527, opened 2024-03-07
feat: add go-format example target

odsod:
Runs `golangci-lint --fix` and `golines`.

### DIFF
--- a/.sage/main.go
+++ b/.sage/main.go
@@ -10,6 +10,7 @@ import (
 	"go.einride.tech/sage/tools/sggo"
 	"go.einride.tech/sage/tools/sggolangcilint"
 	"go.einride.tech/sage/tools/sggolicenses"
+	"go.einride.tech/sage/tools/sggolines"
 	"go.einride.tech/sage/tools/sggoreview"
 	"go.einride.tech/sage/tools/sgmdformat"
 	"go.einride.tech/sage/tools/sgyamlfmt"
@@ -49,6 +50,16 @@ func GoReview(ctx context.Context) error {
 func GoLint(ctx context.Context) error {
 	sg.Logger(ctx).Println("linting Go files...")
 	return sggolangcilint.Run(ctx)
+}
+
+func GoLintFix(ctx context.Context) error {
+	sg.Logger(ctx).Println("fixing Go files...")
+	return sggolangcilint.Fix(ctx)
+}
+
+func GoFormat(ctx context.Context) error {
+	sg.Logger(ctx).Println("formatting Go files...")
+	return sggolines.Run(ctx)
 }
 
 func GoLicenses(ctx context.Context) error {
--- a/Makefile
+++ b/Makefile
@@ -67,6 +67,10 @@ format-yaml: $(sagefile)
 git-verify-no-diff: $(sagefile)
 	@$(sagefile) GitVerifyNoDiff
 
+.PHONY: go-format
+go-format: $(sagefile)
+	@$(sagefile) GoFormat
+
 .PHONY: go-licenses
 go-licenses: $(sagefile)
 	@$(sagefile) GoLicenses
@@ -74,6 +78,10 @@ go-licenses: $(sagefile)
 .PHONY: go-lint
 go-lint: $(sagefile)
 	@$(sagefile) GoLint
+
+.PHONY: go-lint-fix
+go-lint-fix: $(sagefile)
+	@$(sagefile) GoLintFix
 
 .PHONY: go-mod-tidy
 go-mod-tidy: $(sagefile)
--- a/example/.sage/main.go
+++ b/example/.sage/main.go
@@ -10,6 +10,7 @@ import (
 	"go.einride.tech/sage/tools/sggo"
 	"go.einride.tech/sage/tools/sggolangcilint"
 	"go.einride.tech/sage/tools/sggolicenses"
+	"go.einride.tech/sage/tools/sggolines"
 	"go.einride.tech/sage/tools/sggoreview"
 	"go.einride.tech/sage/tools/sgmdformat"
 	"go.einride.tech/sage/tools/sgyamlfmt"
@@ -52,6 +53,16 @@ func GoReview(ctx context.Context) error {
 func GoLint(ctx context.Context) error {
 	sg.Logger(ctx).Println("linting Go files...")
 	return sggolangcilint.Run(ctx)
+}
+
+func GoLintFix(ctx context.Context) error {
+	sg.Logger(ctx).Println("fixing Go files...")
+	return sggolangcilint.Fix(ctx)
+}
+
+func GoFormat(ctx context.Context) error {
+	sg.Logger(ctx).Println("formatting Go files...")
+	return sggolines.Run(ctx)
 }
 
 func GoLicenses(ctx context.Context) error {
--- a/tools/sggofumpt/tools.go
+++ b/tools/sggofumpt/tools.go
@@ -1,4 +1,4 @@
-package sggolines
+package sggofumpt
 
 import (
 	"context"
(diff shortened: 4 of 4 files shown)


.sage/main.go
```diff
@@ -51,6 +52,22 @@ func GoLint(ctx context.Context) error {
...
+func GoFix(ctx context.Context) error {
+	sg.Logger(ctx).Println("fixing Go files...")
+	return sggolangcilint.Fix(ctx)
+}
+
+func GoLines(ctx context.Context) error {
+	sg.Logger(ctx).Println("wrapping Go lines...")
+	return sggolines.Run(ctx)
+}
+
```
alethenorio, line 64: Do we need these 2? If the goal is to have a GoFormat, maybe inline these 2?

Also `go fix` is an [actual command](https://pkg.go.dev/cmd/fix) in the Go cli so if we keep it, we should avoid some confusion and call it something different.

.sage/main.go
```diff
@@ -51,6 +52,22 @@ func GoLint(ctx context.Context) error {
...
+func GoFormat(ctx context.Context) error {
+	sg.Logger(ctx).Println("formatting Go files...")
+	sg.SerialDeps(ctx, GoFix, GoLines)
+	return nil
+}
```
alethenorio, line 69: What about gofmt?

.sage/main.go
fredrikaverpil: I'm thinking it _might_be useful to have easy access to run them separately in the case of formatting conflicts, but I also think it is `GoFormat` we are primarily looking for, so inlining sounds good to me.

.sage/main.go
fredrikaverpil: I'll let @odsod answer but isn't `gofumpt` that, and more?

alethenorio: Ah didn't realize the golangci-lint fix target was running gofumpt. Yeah I think it covers gofmt as well

.sage/main.go
odsod: Sure we can inline

alethenorio: Great. My thinking here is that we will rarely, if ever, want to run one or the other separately so let's start by inlining and avoid bloating the default setup even more if the primary target is to run both together

fredrikaverpil: 💯 agree

fredrikaverpil: I just realized something. Combining these two are not necessary "formatting". For example, we recently added the [`protogetter`](https://github.com/ghostiam/protogetter) linter which has auto-fix capabilities.

Maybe `GoFormat` or `make go-format` is a little bit misleading, as I understand that protogetter changes the behavior, not just formatting.

thall: I think its good to have them separately, so the user can choose, but also supply one target to run all formatters, but if you dont agree you can inline.

fredrikaverpil: > The default config that gets generated is meant to be a batteries included setup for the most basic Go repos so I think it makes more sense here to inline following Go convention of 1 way of doing things.

To avoid confusion, maybe let's not call it `GoFormat` then, as not only formatting is (and likely will continue to be) applied when executing it?

alethenorio: Any suggestions on a better name? Could potentially be called `LinterFix`?

fredrikaverpil: I think let's go with `LinterFix` or `LintFix` (only because it's shorter 😄), what do you think @odsod ?

By not prefixing with `Go`, does it imply that this could in the future be extended with e.g. Python linting/formatting?

odsod: We already have `go-mod-tidy`, `go-test`, `go-lint`, `go-review`, `go-licenses` and more. Why not `go-fix`?

odsod: Since we are talking about inlining anyway, the only target we'll add here is `go-format` so this discussion is moot?

fredrikaverpil: > this discussion is moot?

Well, the protogetter linter will change _behavior_, not formatting. So I still don't think calling this `go-format` is correct.

fredrikaverpil: I think I'll leave the naming up to you to decide 😄

tools/sggolines/tools.go
```diff
@@ -10,13 +10,28 @@ import (
...
 	name    = "golines"
 	version = "0.12.2"
 )
 
+// Run golines on all Go files in the current git root with gofumpt as default formatter.
+func Run(ctx context.Context) error {
+	sg.Deps(ctx, sggofumpt.PrepareCommand)
+	return Command(
+		ctx,
+		"--base-formatter=gofumpt",
```
fredrikaverpil, line 26: Hm, maybe relevant: https://github.com/segmentio/golines/issues/100
Didn't have time to read it all through. I just remember seeing this ...

fredrikaverpil: We can always run with what you've done now - and just keep this issue in mind, in case we see something like that later.

.sage/main.go
```diff
@@ -51,6 +52,16 @@ func GoLint(ctx context.Context) error {
 	return sggolangcilint.Run(ctx)
 }
 
+func GoLintFix(ctx context.Context) error {
+	sg.Logger(ctx).Println("fixing Go files...")
+	return sggolangcilint.Fix(ctx)
```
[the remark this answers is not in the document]
odsod: OK, you're saying we should remove the fix-specific config?

tools/sggolines/tools.go
odsod: Ohh nice find! I'll update

.sage/main.go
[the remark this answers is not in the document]
fredrikaverpil: Yes. 😄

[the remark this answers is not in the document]
odsod: I can do that - I can override to serial mode with a flag instead - but technically can be done in a follow-up PR?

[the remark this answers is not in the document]
fredrikaverpil: It makes sense that `make go-lint` and `make go-lint-fix` uses the same config. Right?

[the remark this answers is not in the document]
odsod: The point of the fix-specific config is to not run the non-fixing linters and speed up the run, and to apply serial mode instead of parallell mode - but probably better to have one single config like you say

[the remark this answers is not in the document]
fredrikaverpil: Ohh, the serial mode. Right!
Sure, we can do that as separate PR.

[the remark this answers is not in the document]
fredrikaverpil: I wonder if you can remove that particular part from the config alltogether and instead specify parallel vs serial on as CLI args...?

odsod:
Having a bit of trouble pushing updates from my train WiFi - a final set of changes to the example file is coming as soon as connectitivy permits...

.sage/main.go
```diff
@@ -51,6 +52,16 @@ func GoLint(ctx context.Context) error {
...
+}
+
+func GoFormat(ctx context.Context) error {
+	sg.Logger(ctx).Println("formatting Go files...")
+	return sggolines.Run(ctx)
+}
```
fredrikaverpil, line 63: Hmmm;

```bash
❯ make go-format     
[go-format] formatting Go files...
[sggofumpt:prepare-command] fetching https://github.com/mvdan/gofumpt/releases/download/v0.6.0/gofumpt_v0.6.0
_darwin_arm64 ...
[sggolines:prepare-command] fetching https://github.com/segmentio/golines/releases/download/v0.12.2/golines_0
.12.2_darwin_all.tar.gz ...
[go-format] [2024-03-07 18:26:18] FATAL Error formatting source: exec: "'gofumpt": executable file not found 
in $PATH
[go-format] exit status 1
make: *** [go-format] Error 1
```

odsod: interesting - that "fix" did look too good to be true!

odsod: except for in edge cases it should work with `gofumpt` so let's start with that?

fredrikaverpil:
```bash
❯ make go-format
[go-format] formatting Go files...
```

👍

fredrikaverpil (approved):
LGTM! 🚀

Let's not forget to take care of the two yaml files next...

thall (approved):

PR #528, opened 2024-03-08
Use one configci.yml file

fredrikaverpil:
### Why?

We have two golangci-lint yaml config files, only because one is expected to run the linters in serial (for auto-fixing) vs the other in parallel (for checking).

### What?

- Remove the `golangci.fix.yml`.
- Have both `Run` and `Fix` use the `golangci.yml` file.
- Specify parallel vs serial as CLI argument.

$${\color{green}+3 \space \color{red}-66}$$


odsod (approved):
Lovely stuff! :broom:

PR #529, opened 2024-03-11
Add biome as a json formatter

joaoeinride:
Ideally the binary should be named `biome` rather than `biome-darwin-arm64` to support other platforms

### DIFF
--- a/tools/sgbiome/tools.go
+++ b/tools/sgbiome/tools.go
@@ -1,0 +1,64 @@
+package sgbiome
+
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+	"runtime"
+
+	"go.einride.tech/sage/sg"
+	"go.einride.tech/sage/sgtool"
+)
+
+const (
+	toolName = "biome"
+	version  = "1.6.0"
+)
+
+func Format(ctx context.Context, paths ...string) error {
+	sg.Deps(ctx, prepareCommand)
+	args := []string{
+		"format",
+		"--write",
+		"--log-kind",
+		"compact",
+	}
+	binDir := sg.FromToolsDir(toolName, version)
+	execDir := filepath.Join(binDir, toolName)
+	args = append(args, paths...)
+
+	if err := sg.Command(
+		ctx,
+		execDir,
+		args...,
+	).Run(); err != nil {
+		return fmt.Errorf("running biome %v", err)
+	}
+
+	return nil
+}
+
+func prepareCommand(ctx context.Context) error {
+	binDir := sg.FromToolsDir(toolName, version)
+	arch := runtime.GOARCH
+	if arch == "amd64" {
+		arch = "x64"
+	}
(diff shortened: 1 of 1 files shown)


tools/sgbiome/tools.go
```diff
@@ -0,0 +1,60 @@
...
+	"context"
+	"fmt"
+	"path/filepath"
+
+	"go.einride.tech/sage/sg"
+	"go.einride.tech/sage/sgtool"
+)
+
+const (
+	toolName = "biome-darwin-arm64"
```
odsod, line 13: should be `biome`

odsod (changes requested):
Need to improve the handling of architectures - see other tools for examples

tools/sgbiome/tools.go
```diff
@@ -0,0 +1,63 @@
...
+	sg.Deps(ctx, prepareCommand)
+
+	args := []string{
+		"format",
+		"--write",
+		"--log-kind",
+		"compact",
+	}
+	binDir := sg.FromToolsDir(toolName, version)
+	execDir := filepath.Join(binDir, "biome-linux-x64")
```
thall, line 28: what if this is downloaded on a mac, then it wont be linux, should follow the format they have for different platforms
image

joaoeinride: ah, I thought I had fixed that. fixing fixing

thall: I think you can copy-paste most from this setup, https://github.com/einride/sage/blob/master/tools/sggolines/tools.go

tools/sgbiome/tools.go
```diff
@@ -0,0 +1,64 @@
...
+	"path/filepath"
+	"runtime"
+
+	"go.einride.tech/sage/sg"
+	"go.einride.tech/sage/sgtool"
+)
+
+const (
+	toolName = "biome"
+	version  = "v1.6.0"
```
thall, line 15: just checking other tools, we seem to write version without `v` prefix

thall (approved):
LGTM

tools/sgbiome/tools.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+const (
+	toolName = "biome"
+	version  = "1.6.0"
+)
+
+func Format(ctx context.Context, paths ...string) error {
+	sg.Deps(ctx, prepareCommand)
+
```
odsod, line 20: nit: avoid adding blank lines in small functions since where to put them is based on individual taste

odsod (approved):
Good stuff! Please consider the style nit before merging

tools/sgbiome/tools.go
odsod: I always comment on this because different people have such different preferences around blank lines that the easiest thing is to keep them to a minimum (at least in small-ish functions).

PR #543, opened 2024-05-17
unexpected memoization of sage targets

thall:
Sage targets results are memoized, so if invoked multiple times, the result from the first execution is returned.

In this example PR, `GitVeryifyNoDiff` is called first, then `FormatMarkdown` and then `GitVeryifyNoDiff`, then second call wont report any changes since it gets a cached value back, to me this was unexpected.

The memoization happens [here](https://github.com/einride/sage/blob/master/sg/internal/runner/runner.go#L17)

The behavior is verified with a exaction of `git status --short`:
image

### DIFF
--- a/.sage/main.go
+++ b/.sage/main.go
@@ -24,6 +24,13 @@ func main() {
 	)
 }
 
+func DoubleGitVerifyNoDiff(ctx context.Context) error {
+	sg.SerialDeps(ctx, GitVerifyNoDiff)
+	sg.SerialDeps(ctx, FormatMarkdown)
+	sg.SerialDeps(ctx, GitVerifyNoDiff) // this should report a change
+	return nil
+}
+
 func Default(ctx context.Context) error {
 	sg.Deps(ctx, ConvcoCheck, GoLint, GoTest, FormatMarkdown, FormatYaml, BackstageValidate)
 	sg.SerialDeps(ctx, GoModTidy)
--- a/.sage/main.go
+++ b/.sage/main.go
@@ -32,10 +32,9 @@ func DoubleGitVerifyNoDiff(ctx context.Context) error {
 }
 
 func Default(ctx context.Context) error {
-	sg.Deps(ctx, ConvcoCheck, GoLint, GoTest, FormatMarkdown, FormatYaml, BackstageValidate)
-	sg.SerialDeps(ctx, GoModTidy)
-	sg.SerialDeps(ctx, GoLicenses, GitVerifyNoDiff)
-	return nil
+	sg.Deps(ctx, DoubleGitVerifyNoDiff)
+	cmd := sggit.Command(ctx, "status", "--short")
+	return cmd.Run()
 }
 
 func GoModTidy(ctx context.Context) error {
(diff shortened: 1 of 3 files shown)


.sage/main.go
```diff
@@ -24,6 +24,13 @@ func main() {
 	)
 }
 
+func DoubleGitCheck(ctx context.Context) error {
+	sg.SerialDeps(ctx, GitVerifyNoDiff)
+	sg.SerialDeps(ctx, GoFormat)
+	sg.SerialDeps(ctx, GitVerifyNoDiff) // this should report a change
```
thall, line 30: Unexpectedly this call does not return an error, even though the previous format changed a file

.sage/main.go
```diff
@@ -32,10 +32,9 @@ func DoubleGitCheck(ctx context.Context) error {
 }
 
 func Default(ctx context.Context) error {
-	sg.Deps(ctx, ConvcoCheck, GoLint, GoTest, FormatMarkdown, FormatYaml, BackstageValidate)
-	sg.SerialDeps(ctx, GoModTidy)
-	sg.SerialDeps(ctx, GoLicenses, GitVerifyNoDiff)
-	return nil
+	sg.Deps(ctx, DoubleGitCheck)
+	cmd := sggit.Command(ctx, "status", "--short")
```
thall, line 36: `DoubleGitCheck` wont return any changes, but `git status --short` verifies that files actually were changed.

.sage/main.go
```diff
@@ -32,10 +32,9 @@ func DoubleGitVerifyNoDiff(ctx context.Context) error {
 }
 
 func Default(ctx context.Context) error {
-	sg.Deps(ctx, ConvcoCheck, GoLint, GoTest, FormatMarkdown, FormatYaml, BackstageValidate)
-	sg.SerialDeps(ctx, GoModTidy)
-	sg.SerialDeps(ctx, GoLicenses, GitVerifyNoDiff)
-	return nil
+	sg.Deps(ctx, DoubleGitVerifyNoDiff)
+	cmd := sggit.Command(ctx, "status", "--short")
```
thall, line 36: `DoubleGitVerifyNoDiff` wont fail even do a file were modified, but `git status --short` verifies that a file actually were changed.

README.md
```diff
@@ -6,7 +6,6 @@ provides a curated and maintained set of build [tools](./tools) for Go projects.
 [![Release](https://github.com/einride/sage/actions/workflows/release.yml/badge.svg)](https://github.com/einride/sage/actions/workflows/release.yml)
 
 ## Requirements
-
```
thall, line 9: This incorrect formatting will be fixed by the GHA job, leading to a dirty workspace, but the job will still be OK

.sage/main.go
```diff
@@ -24,6 +24,13 @@ func main() {
 	)
 }
 
+func DoubleGitVerifyNoDiff(ctx context.Context) error {
+	sg.SerialDeps(ctx, GitVerifyNoDiff)
+	sg.SerialDeps(ctx, FormatMarkdown)
+	sg.SerialDeps(ctx, GitVerifyNoDiff) // this should report a change
```
thall, line 30: Unexpectedly the second call to `GitVerifyNoDiff` does not return an error, even though the previous format changed a file

PR #544, opened 2024-05-20
feat(cco-1025): add setup for local test postgres instance to support using a shared test instance

sharkema:
This pr will add a tool 'sgpostgres' tool similar to 'sgspanner' emulator, but then for Postgres. 

On the start of a test run (`make go-test`), sage can spin up a local postgres instance in a docker container. This db instance can be shared between tests, such that not every test needs to spin up a separate database instance. Similar to sgspanner, it posts the address how to connect to the db in an environment property.

For charge-control this will improve test run times by ~40%.

### DIFF
--- a/tools/sgpostgres/local.go
+++ b/tools/sgpostgres/local.go
@@ -1,0 +1,165 @@
+package sgpostgres
+
+import (
+	"bytes"
+	"context"
+	"fmt"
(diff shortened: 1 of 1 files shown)


odsod (approved):
Good stuff - LGTM

PR #545, opened 2024-05-31
feat(sggolicenses): add ignored dependencies

hug-dev:
Adds two go-licenses check functions to be able to pass repositories to be ignored. This might be needed if the repositories have an open-source license in a format which is not recognized by the tool.

See for example: https://github.com/ajstarks/svgo/issues/73 and https://github.com/golang/freetype/issues/94

### DIFF
--- a/tools/sggolicenses/tools.go
+++ b/tools/sggolicenses/tools.go
@@ -25,9 +25,7 @@ func Command(ctx context.Context, args ...string) *exec.Cmd {
 	return sg.Command(ctx, sg.FromBinDir(name), args...)
 }
 
-// Check for disallowed types of Go licenses in a specific directory.
-// By default, Google's forbidden and restricted types are disallowed.
-func CheckDir(ctx context.Context, directory string, disallowedTypes ...string) error {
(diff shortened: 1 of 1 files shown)


tools/sggolicenses/tools.go
```diff
@@ -57,9 +53,45 @@ func CheckDir(ctx context.Context, directory string, disallowedTypes ...string)
...
+		appendArgs = append(appendArgs, "--disallowed_types=forbidden,restricted")
+	}
+
+	return checkDir(ctx, directory, appendArgs)
+}
+
+// Check for disallowed types of Go licenses in a specific directory.
+// By default, Google's forbidden and restricted types are disallowed.
+// Allows to ignore some dependencies by specifying their package path (ie "github.com/einride/sage").
+func CheckDirWithIgnored(
```
odsod, line 72: with this many helpers it might be better to inline it in the specific repos where it's used - the code here is just meant to be the "most common default" use case - not all possible use cases

hug-dev: will close for now, the problem in the repo has been solved in another way

PR #556, opened 2024-07-09
feat(sgcloudrun): add LocalDevelop target and corresponding command

alethenorio:
Add a new LocalDevelop target which runs a cloud run service
locally using an impersonated service account which gives the benfit of
using short-lived tokens for better security and avoids having to refer
to a service account key.

This deprecates the old targets as the new ones are essentially a
better, more secure, replacement.

### DIFF
--- a/tools/sgcloudrun/config.go
+++ b/tools/sgcloudrun/config.go
@@ -5,10 +5,13 @@ import (
 	"context"
 	"encoding/base64"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"io"
 	"net/http"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
 
 	"go.einride.tech/sage/sg"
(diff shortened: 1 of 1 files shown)


tools/sgcloudrun/config.go
```diff
@@ -28,6 +33,8 @@ func Develop(ctx context.Context, path, keyFile, configFile string) error {
 
 // DevelopCommand returns an *exec.Cmd pre-configured to start the Cloud Run service at the provided Go path
 // with the provided service account and config.
+// Deprecated: DevelopCmmand uses a service account key which are inherently more risky as they are not often rotated.
```
odsod, line 36: DevelopCmmand => DevelopCommand

tools/sgcloudrun/config.go
```diff
@@ -219,3 +300,73 @@ func accessSecretVersion(ctx context.Context, accessToken, project, secret, vers
 	}
 	return string(decodedData), nil
 }
+
+func applicationDefaultCredential() (string, error) {
```
odsod, line 304: applicationDefaultCredentials* (they use plural in Google's APIs)

odsod (approved):
LGTM - if this has been tested in a few services, let's get this show on the road! :rocket:

tools/sgcloudrun/config.go
alethenorio: Fixed

tools/sgcloudrun/config.go
alethenorio: Fixed

PR #563, opened 2024-08-05
feat(sg/path): use os cache if possible

ngalaiko:


odsod:
@ngalaiko this adds a fair bit of complexity without making the CI/CD builds faster - is it really worth it?

ngalaiko:
> @ngalaiko this adds a fair bit of complexity without making the CI/CD builds faster - is it really worth it?

I was mostly thinking about local development here. It's quite common for such tools to share resources.

About complexity, I don't agree it's much. The only source of complexity here is duplication of logic to resolve os dependant paths in both `go` and `make`.

What could be problematic is that Makefiles will change after every `sage` upgrade, because now they include sage version. 

Feel free to discard it anyways, it's really not that big of a deal.

sg/generate.go
```diff
@@ -76,3 +80,19 @@ func GenerateMakefiles(mks ...Makefile) {
...
+		if stat.Mode().Type() == os.ModeSymlink {
+			return
+		}
+		if removeErr := os.RemoveAll(newname); removeErr != nil {
+			panic(removeErr)
+		}
+	}
+	ensureParentDir(newname)
+	if err := os.Symlink(oldname, newname); err != nil {
+		panic(fmt.Errorf("error creating symlink: %w", err))
```
ccoVeille, line 96: Maybe adding the filename in the error could help to debug

Something like


```suggestion
		panic(fmt.Errorf("error creating symlink %q: %w", newname, err))
```

sg/path.go
```diff
@@ -24,6 +26,34 @@ func FromWorkDir(pathElems ...string) string {
...
+	case "linux":
+		if xdgCacheHome := os.Getenv("XDG_CACHE_HOME"); xdgCacheHome != "" {
+			return filepath.Join(append([]string{xdgCacheHome}, pathElems...)...), true
+		} else {
+			return filepath.Join(append([]string{os.Getenv("HOME"), ".cache"}, pathElems...)...), true
+		}
+	default:
+		return "", false
+	}
+}
```
ccoVeille, line 44: Is there a problem with [os.UserCacheDir](https://pkg.go.dev/os#UserCacheDir) ?

If so you should document it in a comment

sg/path.go
```diff
@@ -24,6 +26,34 @@ func FromWorkDir(pathElems ...string) string {
 	return filepath.Join(append([]string{cwd}, pathElems...)...)
 }
 
+// FromOSCache reutrns the path relative to OS cache folder, and a boolean indicating
```
ccoVeille, line 29:
```suggestion
// FromOSCache returns the path relative to OS cache folder, and a boolean indicating
```

sg/path.go
```diff
@@ -24,6 +26,34 @@ func FromWorkDir(pathElems ...string) string {
...
+// whether such folder exists or not.
```
ccoVeille, line 30:
```suggestion
// whether such folder exists.
```

When using "whether", you should not add "or not" as it's implied

sg/path.go
```diff
@@ -24,6 +26,34 @@ func FromWorkDir(pathElems ...string) string {
...
+func FromOSSageCache(pathElems ...string) (string, bool) {
+	if version, isSageVersionSet := GetSageVersion(); isSageVersionSet {
+		osCacheDir, isOSCacheExists := FromOSCache([]string{modulePath, version}...)
+		if err := os.MkdirAll(osCacheDir, 0o700); err != nil {
+			panic(err)
+		}
+		return filepath.Join(append([]string{osCacheDir}, pathElems...)...), isOSCacheExists
+	}
+	return "", false
+}
```
ccoVeille, line 55:
```suggestion
func FromOSSageCache(pathElems ...string) (string, bool) {
	version, isSageVersionSet := GetSageVersion()
	if !isSageVersionSet {
		return "", false
	}
	osCacheDir, isOSCacheExists := FromOSCache([]string{modulePath, version}...)
	if err := os.MkdirAll(osCacheDir, 0o700); err != nil {
		panic(err)
	}
	return filepath.Join(append([]string{osCacheDir}, pathElems...)...), isOSCacheExists
}
```

ccoVeille (changes requested):

ngalaiko:
thanks @ccoVeille, useful feedback

thall:
Will `make clean-sage` remove all dependencies for all repos with this PR?

ngalaiko:
> Will `make clean-sage` clean the central cache with this PR?

I'll remove symlinks from git with this pr. Similar to what it did before. `git rm`, not actual `rm`

thall:
> > Will `make clean-sage` clean the central cache with this PR?
> 
> I'll remove symlinks from git with this pr. Similar to what it did before. `git rm`, not actual `rm`

Great!

ngalaiko:
Before merging, make sure that cache path is correct.

PR #571, opened 2024-08-15
fix: address unreliable go dependency caching

fredrikaverpil:
## Why this change?

This aims to solve the problem where two caches are uploaded (one by `actions/go-setup` and one by `actions/cache`, but where the one from `actions/go-setup` was created by the fastest workflow (with the least
dependencies) and then often used by the slower/heavier workflows. These heavier workflows reuse the cache but it does not contain all dependencies and so it has to re-download and re-build them on every commit. The cache is not rebuilt because there are additional dependencies downloaded.

TL;DR the wrong cache is being used.

The issue is described in detail here: https://github.com/actions/setup-go/issues/358

## What was changed?

- Stop using cache by `actions/go-setup`.
- Extend caching of `actions/cache` so to also cover Go dependencies and builds. Effectively doing what `actions/go-setup` was supposed to handle.
- Extend the cache key to use a hash based on _all_ `go.sum` files in the project, not just `.sage/go.sum`.

### DIFF
--- a/actions/setup/action.yml
+++ b/actions/setup/action.yml
@@ -53,7 +53,7 @@ runs:
           /home/runner/.cache/go-build
           /home/runner/go/pkg/mod
           /home/runner/go/bin
-        key: ${{ runner.os }}-sage-${{ github.workflow }}-${{ github.job }}-${{ inputs.cacheKey }}-${{ inputs.go-version }}-${{ hashFiles('./.sage/go.sum') }}
+        key: ${{ runner.os }}-sage-${{ github.workflow }}-${{ github.job }}-${{ inputs.cacheKey }}-${{ inputs.go-version }}-${{ hashFiles('**/go.sum') }}
         restore-keys: |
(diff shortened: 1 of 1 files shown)


actions/setup/action.yml
```diff
@@ -56,7 +57,10 @@ runs:
         path: |
           ./.sage/tools
           ./.sage/bin
-        key: ${{ runner.os }}-sage-${{ github.workflow }}-${{ github.job }}-${{ inputs.cacheKey }}-${{ inputs.go-version }}-${{ hashFiles('./.sage/go.sum') }}
+          /home/runner/.cache/go-build
+          /home/runner/go/pkg/mod
+          /home/runner/go/bin
+        key: ${{ runner.os }}-sage-${{ github.workflow }}-${{ github.job }}-${{ inputs.cacheKey }}-${{ inputs.go-version }}-${{ hashFiles('**/go.sum') }}
```
fredrikaverpil, line 63: This change is not needed for improving the go caching, but I figured it would be an overall improvement and included it here as a bonus 😄

alethenorio (approved):

PR #574, opened 2024-08-20
feat(spanner-aip): add to tools

ngalaiko:


tools/sgspanneraipgo/tools.go
```diff
@@ -0,0 +1,24 @@
...
+	name    = "spanner-aip"
+	version = "v0.61.0"
+)
+
+func Command(ctx context.Context, args ...string) *exec.Cmd {
+	sg.Deps(ctx, PrepareCommand)
+	return sg.Command(ctx, sg.FromBinDir(name), args...)
+}
+
+func PrepareCommand(ctx context.Context) error {
```
odsod, line 21: you generally want to do this with the version that's specified in your `go.mod` to make sure you have compatible runtime with code generated APIs

odsod: not sure it adds enough value to bring this in here, given that it'll lead to code generator and runtime falling out of sync in places that use this helper

ngalaiko: aha, so that's the reason. I was about to write an explanation, see in slack

PR #577, opened 2024-08-23
fix(gha): allow opt-out of **/go.sum repo traversal

fredrikaverpil:
## Why this change?

We have seen issues where the runner does not have permissions to enter a
certain directory. Then the post-step of the cache will fail in GHA.

The reason is that `hashfiles` currently traverses the entire repository.

## What was changed?

- Move the `hashfiles` entry, which is part of the cache key to become opt-out.

## Notes

- This PR supersedes #576

### DIFF
--- a/actions/setup/action.yml
+++ b/actions/setup/action.yml
@@ -2,11 +2,6 @@ name: Setup Sage
 description: Setup Sage, including caching of tools and Go modules.
 
 inputs:
-  cacheKey:
-    description: Custom cache key used
-    required: false
-    default: cachekey
-
   disableCache:
     description: Disable cache
     required: false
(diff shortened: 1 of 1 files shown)


actions/setup/action.yml
```diff
@@ -17,6 +12,13 @@ inputs:
     required: false
     default: '1.22'
 
+  cacheKey:
```
alethenorio, line 15: Is this a cache prefix rather than a cache key?

actions/setup/action.yml
```diff
@@ -17,6 +12,13 @@ inputs:
...
+    description: Custom cache key used
+    required: false
+    # NOTE: some projects cannot use **/go.sum because: https://github.com/actions/runner/issues/449
+    # In such cases, try providing the cache key and set it to e.g. ${{ inputs.go-version }}-${{ hashFiles('go.sum', '*/go.sum') }}
+    default: -${{ hashFiles('**/go.sum') }}
```
alethenorio, line 20: The value this is appended to already contains a dash so it would cause a double dash :)

fredrikaverpil: Whoops 😄 fixed!

actions/setup/action.yml
fredrikaverpil: Yes, but we already had this input since before so I didn't want to rename it.

fredrikaverpil: You can see here where `cacheKey` is used: https://github.com/search?q=org%3Aeinride+cacheKey+language%3AYAML+path%3A%2F%5E%5C.github%5C%2Fworkflows%5C%2F%2F&type=code

fredrikaverpil:
I'm closing this and we can revive it if needed.

PR #580, opened 2024-08-28
feat(golangcilint): replace exportloopref with copyloopvar

christka1:
We currently get the following warning when running the linter:
"The linter 'exportloopref' is deprecated (since v1.60.2) due to:
Since Go1.22 (loopvar) this linter is no longer relevant. Replaced by copyloopvar."


odsod (approved):

radhus (approved):
I would expect to see some linter errors with this, as you now need to cleanup the previous loop var workaround. Good stuff 👍

PR #582, opened 2024-09-05
feat: tools/golangci-lint disable G115

alethenorio:
Prone to false positives.

### DIFF
--- a/tools/sggolangcilint/golangci.yml
+++ b/tools/sggolangcilint/golangci.yml
@@ -1,6 +1,12 @@
 run:
   timeout: 10m
 
+linter-settings:
+  gosec:
+    excludes:
+      # Flags for potentially-unsafe casting of ints. Prone to lots of false positives.
+      - G115
+
 issues:
   fix: false
   exclude-dirs:


SCKelemen (approved):

parmus (approved):
We could even add a comment with
`https://github.com/securego/gosec/issues?q=is%3Aissue+G115` so it's easy to follow existing issues and if they have been addressed.

odsod (approved):

ccoVeille:
gosec was improved, there were indeed a lot false positive.

A lot of them has been addressed in gosec 2.1.2

golangci-lint merged them to master.

I would recommend waiting the next golangci-lint version before disabling G115

parmus:
> gosec was improved, there were indeed a lot false positive.
> 
> A lot of them has been addressed in gosec 2.21.1
> 
> golangci-lint merged them to master today, but they are not yet released
> 
> I would recommend waiting the next golangci-lint version before disabling G115

Even if gosec has been improved, it doesn't change the fact that golangci-lint hasn't been updated, and there is no indication on when that might happen. This means that even if I address the issue in my repo with a helper function like this:
```
func safeConvert(value int) (int32, error) {
	if value < math.MinInt32 || value > math.MaxInt32 {
		return 0, fmt.Errorf("Int32 overflow: %v", value)
	}
	// #nosec G115 -- gosec is still too stupid to recognize that the above it a valid range check.
	return int32(value), nil
}
```
I still need to add `// #nosec G115` because the range check isn't being recognized. And I have to add at least one of these in basically every single repo due to past mistakes when designing protos.

Again I'd argue that this is bad, because it leaves at lease one `// #nosec G115` in each of our repo, which someone has to manually remember to go back and clean up later. Has however we disable G115 in Sage, it would be trivial to re-enable it at a later point in time when the gosec implementation has been fixed, and the effect would be company-wide.

ccoVeille:
> ```
> func safeConvert(value int) (int32, error) {
> 	if value < math.MinInt32 || value > math.MaxInt32 {
> 		return 0, fmt.Errorf("Int32 overflow: %v", value)
> 	}
> 	// #nosec G115 -- gosec is still too stupid to recognize that the above it a valid range check.
> 	return int32(value), nil
> }
> ```

BTW, I'm working on a lib to handle this pesty gosec G115

https://github.com/ccoveille/go-safecast

I published the v1.0.0 a bit too fast. I have a few PRs pending, to address the few issues I identified.

alethenorio:
Thank you for the information @ccoVeille 

I'm merging this one to pick it up at a later date when things have improved

parmus:
> BTW, I'm working on a lib to handle this pesty gosec G115
> 
> https://github.com/ccoveille/go-safecast
> 
> I published the v1.0.0 a bit too fast. I have a few PRs pending, to address the few issues I identified.

Great stuff! I just wish I'd seen this before working on my own solution: https://github.com/einride/safecast-go/blob/master/integers.go :sweat_smile: 

Let's combine forces instead and get a single solid solution going. Let me know if you want any assistance. And feel free to steal anything useful from my solution, if there's anything you like.

PR #585, opened 2024-09-11
feat(sgcspevaluatorcli): add csp helper type

ngalaiko:
ContentSecurityPolicy type provides a set of helper functions to define content security policy in a structured way.

### DIFF
--- a/tools/sgcspevaluatorcli/csp.go
+++ b/tools/sgcspevaluatorcli/csp.go
@@ -1,0 +1,80 @@
+package sgcspevaluatorcli
+
+import (
+	"fmt"
+	"sort"
+	"strings"
+)
+
+// ContentSecurityPolicy is a map where each key is a Content Security Policy directive,
+// and value is a list of sources for the respective directive.
+type ContentSecurityPolicy map[string][]string
+
+// String returns ContentSecurityPolicy ready to be send as a HTTP header.
+func (csp ContentSecurityPolicy) String() string {
+	directives := make([]string, 0, len(csp))
+	for directive, rules := range csp {
+		sort.Slice(rules, func(i, j int) bool { return rules[i] < rules[j] })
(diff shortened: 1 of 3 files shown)


tools/sgcspevaluatorcli/csp.go
```diff
@@ -0,0 +1,77 @@
...
+	cloned := map[string][]string{}
+	for key, values := range csp {
+		cloned[key] = make([]string, len(values))
+		copy(cloned[key], values)
+	}
+	return cloned
+}
+
+// With returns a new content security policy with an additional directive and sources.
+func (csp ContentSecurityPolicy) With(directive, source string, sources ...string) ContentSecurityPolicy {
```
odsod, line 35: how come we separate out the first vararg to a separate arg here?

ngalaiko: that's to force at least one source on the compilation level.

i.e.
```
csp.With("directive")
```

won't compile

odsod: ah! I see, nice way of avoiding errors/panics!

tools/sgcspevaluatorcli/csp.go
```diff
@@ -0,0 +1,80 @@
...
+	for key, values := range csp {
+		cloned[key] = make([]string, len(values))
+		copy(cloned[key], values)
+	}
+	return cloned
+}
+
+// With returns a new content security policy with an additional directive and sources.
+//
+// Note that the _new_ policy is returned here, so changing the returned value from this function will
```
ngalaiko, line 36: @radhus i've added this note. does this make it clear why cloning is there?

radhus: Yes 👍

radhus (approved):

odsod (approved):
Good stuff!

PR #586, opened 2024-09-11
feat: update golangcilint to v1.61.0

mlveggo:


ccoVeille (commented):
Any plan to reactivate gosec G115?

Most of the false-positive has been addressed by gosec, and were added to golangci-lint 1.61.0

quoral (approved):

PR #587, opened 2024-09-16
sgcloudrun: avoid creds.json conflict

ngalaiko:
if two processes run LocalDevelopment command, they should use
different credes.json to avoid conflict.

also adds support for context cancellation when CTRL-C detected.

### DIFF
--- a/sg/initfile.go
+++ b/sg/initfile.go
@@ -49,6 +49,18 @@ func generateInitFile(g *codegen.File, pkg *doc.Package, mks []Makefile) error {
 		}
 		g.P("logger := ", g.Import("go.einride.tech/sage/sg"), ".NewLogger(\"", loggerName, "\")")
 		g.P("ctx = ", g.Import("go.einride.tech/sage/sg"), ".WithLogger(ctx, logger)")
+		g.P("ctx, cancel := context.WithCancel(ctx)")
+		g.P("defer cancel()")
+
+		g.P("go func() {")
+		g.P("shutdownCh := make(chan ", g.Import("os"), ".Signal, 1)")
+		g.P(g.Import("os/signal"), ".Notify(shutdownCh, ",
+			g.Import("os"), ".Interrupt, ",
+			g.Import("syscall"), ".SIGTERM)",
+		)
+		g.P("<-shutdownCh")
+		g.P("cancel()")
+		g.P("}()")
 		if len(function.Decl.Type.Params.List) > 1 {
 			expected := countParams(function.Decl.Type.Params.List) - 1
 			g.P("if len(args) != ", expected, " {")
--- a/tools/sgcloudrun/config.go
+++ b/tools/sgcloudrun/config.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math/rand/v2"
 	"net/http"
 	"os"
 	"os/exec"
@@ -128,7 +129,7 @@ func LocalDevelopCommand(
 	if err := os.MkdirAll(workDir, 0o755); err != nil {
 		return nil, fmt.Errorf("unable to create path to store gcloud credentials: %v", err)
 	}
-	credsPath := filepath.Join(workDir, "creds.json")
+	credsPath := filepath.Join(workDir, fmt.Sprintf("creds-%s.json", randomLower(5)))
 	if err := os.WriteFile(credsPath, delegateCredsJSON, 0o600); err != nil {
 		return nil, err
 	}
@@ -140,6 +141,12 @@ func LocalDevelopCommand(
 	cmd.Env = append(cmd.Env, "GOOGLE_APPLICATION_CREDENTIALS="+credsPath)
 	cmd.Env = append(cmd.Env, env...)
 	cmd.Env = append(cmd.Env, os.Environ()...) // allow environment overrides
+	cmd.Cancel = func() error {
+		if err := cmd.Process.Kill(); err != nil {
+			return err
+		}
+		return os.Remove(credsPath)
+	}
 	return cmd, nil
 }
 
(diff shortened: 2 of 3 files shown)


sg/initfile.go
```diff
@@ -49,6 +49,20 @@ func generateInitFile(g *codegen.File, pkg *doc.Package, mks []Makefile) error {
...
+
+		g.P("go func() {")
+		g.P("shutdownCh := make(chan ", g.Import("os"), ".Signal, 1)")
+		g.P(g.Import("os/signal"), ".Notify(shutdownCh, ",
+			g.Import("os"), ".Interrupt, ",
+			g.Import("syscall"), ".SIGKILL, ",
+			g.Import("syscall"), ".SIGTERM)",
+		)
+		g.P("sig := <-shutdownCh")
+		g.P("logger.Printf(\"received: %s\", sig)")
```
alethenorio, line 63: Do we need this? Can we remove it?

sg/initfile.go
```diff
@@ -49,6 +49,20 @@ func generateInitFile(g *codegen.File, pkg *doc.Package, mks []Makefile) error {
...
+		g.P("cancel()")
```
alethenorio, line 64: If we are already deferring the cancel call, do we need to call it here as well?

sg/initfile.go
ngalaiko: removed

sg/initfile.go
ngalaiko: because we want to cancel context in two cases:
1. if command finished on it's own. this is done by defering context
2. if we see an os signal. this line handles that. once context is cancelled, it will trigger a shutdown that will trigger all cmd.Cancel functions and kill every child

quoral: Technically this would mean that the context is double cancelled if a Signal terminates us. This is not great, but also has no terrible sideeffects that I've been able to find or discern from a brief test like:
```
func main() {
	_, cancel := context.WithCancel(context.Background())
	cancel()
	cancel()
}
```

If you want to be super correct, observing the context and see whether it needs to be cancelled might be the "best", but seems like a lot of extra work for no benefit as my google-foo can find.

quoral: Also, by checking out the code for how a cancel actually works, it seems indeed to be a safe operation to double cancel. https://cs.opensource.google/go/go/+/refs/tags/go1.23.1:src/context/context.go;l=544-547
```
func (c *cancelCtx) cancel(removeFromParent bool, err, cause error) {
	if err == nil {
		panic("context: internal error: missing cancel error")
	}
	if cause == nil {
		cause = err
	}
	c.mu.Lock()
	if c.err != nil {
		c.mu.Unlock()
		return // already canceled
	}
```

alethenorio: Agreed.

tools/sgcloudrun/config.go
```diff
@@ -370,3 +377,15 @@ func fetchImpersonatedAccessToken(ctx context.Context, serviceAccountEmail strin
...
 	return tokens.AccessToken, nil
 }
+
+//nolint:gochecknoglobals
+var lower = []rune("abcdefghijklmnopqrstuvwxyz")
+
+func randomLower(n uint32) string {
+	b := make([]rune, n)
+	for i := range b {
+		//nolint:gosec
```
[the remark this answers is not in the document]
ngalaiko: because math is pseudo-random. in this case we don't care

sg/initfile.go
ngalaiko: I think it's ok. Cancelling context closes channel, so double cancelling should be fine 🤔

ngalaiko: however i agree, it is weird. i'll refactor this

quoral: Imho, if you find a clean way to check it, do it. Otherwise just ignore it and move on - its a very minor thing and has no sideefects unless they break the Cancel functionality in a future Golang release.

ngalaiko: yea, i'd be surprised if they break it. this would mean closing the same channel twice 🙃

alethenorio (approved):

tools/sgcloudrun/config.go
```diff
@@ -370,3 +377,15 @@ func fetchImpersonatedAccessToken(ctx context.Context, serviceAccountEmail strin
 
 	return tokens.AccessToken, nil
 }
+
+//nolint:gochecknoglobals
+var lower = []rune("abcdefghijklmnopqrstuvwxyz")
```
odsod, line 382: could define a private function returning this and it'll get inlined anyway?

tools/sgcloudrun/config.go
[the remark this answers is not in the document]
odsod: add a comment after nolint directives giving a short explanation of why it's ignored, and all questions will be answered :)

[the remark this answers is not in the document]
odsod: from the docs:

> You may add a comment explaining or justifying why //nolint is being used on the same line as the flag itself:

```go
//nolint:gocyclo // This legacy function is complex but the team too busy to simplify it
func someLegacyFunction() *string {
  // ...
}
```go

File: tools/sgcloudrun/config.go (hunk 4 above)
Comment (ngalaiko) [reply tools/sgcloudrun/config.go]: did this instead:
```
func randomLower(n uint32) string {
	b := make([]rune, n)
	for i := range b {
		// NOTE: code of 'a' is 97, code of 'z' is 122.
		//       so rand.IntN(26) + 97 is code of a "random" lowercase rune.
		//nolint:gosec // we don't need a secure randomizer for this.
		b[i] = rune(rand.IntN(26) + 97)
	}
	return string(b)
}```

tools/sgcloudrun/config.go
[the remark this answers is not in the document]
ngalaiko: done 👍

sg/initfile.go
ngalaiko: didn't find a good way of refactoring that wouldn't introduce more complexity 🤔

PR #588, opened 2024-09-19
GolangCI Lint runs for all targets

joaoeinride:
Before this PR Sage Golangci-Lint would run for all sub folders where a go.mod is found and terminate when the first command returns an errors, could be a lint issue or some other error.
After Golangci-Lint will run for all sub folders and will run for all projects and join all errors together.

### DIFF
--- a/tools/sggolangcilint/tools.go
+++ b/tools/sggolangcilint/tools.go
@@ -63,12 +63,11 @@ func Run(ctx context.Context, args ...string) error {
 	}); err != nil {
 		return err
 	}
+	errs := make([]error, 0, len(commands))
 	for _, cmd := range commands {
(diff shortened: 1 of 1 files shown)


odsod (approved):

fredrikaverpil:
@joaoeinride FYI `errors.Join` was introduced in Go 1.20, I think. And sage is supposed to be compatible with Go 1.17, judging from the `go.mod`:

https://github.com/einride/sage/blob/master/go.mod#L3

I noticed this became a problem as we cannot update Sage in an older project which uses Go 1.19 because of the addition of `errors.Join`. Figured I'd just mention it here.

cc @alethenorio (in case I have misunderstood the versions/dependencies situation here)

PR #592, opened 2024-10-02
feat(sggh): add support for macOS

kutkaite:


tools/sggh/tools.go
```diff
@@ -23,24 +23,39 @@ func Command(ctx context.Context, args ...string) *exec.Cmd {
...
 		version,
 		version,
 		hostOS,
 		hostArch,
+		ext,
 	)
-	if err := sgtool.FromRemote(
-		ctx,
-		binURL,
+
```
odsod, line 42: pls avoid adding blank lines (everyone has different taste for blank lines so to be consistent, try and limit)

odsod (approved):
Please fix the styling comment before merge

tools/sggh/tools.go
kutkaite: sure, I can get rid of it for you 😄

PR #595, opened 2024-10-04
feat(spanner): bump to 1.5.22

fredrikaverpil:
I'm not bumping all the way to 1.5.24 as we are seeing issues with Spanner on
macOS on versions released after 1.5.22.

I've added a note on this limitation/issue in our internal macOS tech-radar.

We should investigate and perhaps file an issue with
https://github.com/GoogleCloudPlatform/cloud-spanner-emulator

### DIFF
--- a/tools/sgcloudspanner/emulator.go
+++ b/tools/sgcloudspanner/emulator.go
@@ -17,7 +17,7 @@ import (
 // https://console.cloud.google.com/gcr/images/cloud-spanner-emulator/global/emulator
 const (
 	url     = "gcr.io/cloud-spanner-emulator/emulator"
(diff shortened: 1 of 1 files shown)


odsod (approved):

odsod:
@fredrikaverpil filing issue sounds great, please do if you have time! :pray:

fredrikaverpil:
@odsod FYI it was not them, it was us 😄 (as expected)
@radhus found the problem, which is that we pass pointers into transactions (which get mutated by e.g. our internal spanner timestamp marker). Spanner is free to retry the transaction - and this got a lot more frequent in spanner emulator 1.5.23 (which is a good thing).

PR #603, opened 2024-10-15
Add protoc-gen-openapi (from gnostic)

fredrikaverpil:
### Why this change?

So that we can generate openapiv3 JSON file from protos using `protoc-gen-openapi` and version control it from Sage.

### Notes

Docs: https://github.com/google/gnostic/tree/main/cmd/protoc-gen-openapi

### DIFF
--- a/tools/sgprotocgenopenapi/tools.go
+++ b/tools/sgprotocgenopenapi/tools.go
@@ -1,0 +1,28 @@
+package sgprotocgenopenapi
+
+import (
+	"context"
+	"os/exec"
+
+	"go.einride.tech/sage/sg"
+	"go.einride.tech/sage/sgtool"
+)
+
+// docs:
+// https://github.com/google/gnostic/tree/main/cmd/protoc-gen-openapi
+// https://buf.build/gnostic/gnostic/docs/main:gnostic.openapi.v3
+
+const (
+	version = "0.7.0"
+	name    = "protoc-gen-openapi"
+)
+
+func Command(ctx context.Context, args ...string) *exec.Cmd {
+	sg.Deps(ctx, PrepareCommand)
+	return sg.Command(ctx, sg.FromBinDir(name), args...)
+}
+
+func PrepareCommand(ctx context.Context) error {
+	_, err := sgtool.GoInstall(ctx, "github.com/google/gnostic/cmd/"+name, "v"+version)
+	return err
+}


radhus (approved):

alethenorio (commented):
Does the package name need to contain the openapi version? The tool seems to be generic so maybe `sgprotocgenopenapi` is enough?

fredrikaverpil:
> Does the package name need to contain the openapi version? The tool seems to be generic so maybe `sgprotocgenopenapi` is enough?

@alethenorio good one, I fixed that.

tools/sgprotocgenopenapi/tools.go
```diff
@@ -0,0 +1,28 @@
...
+	name    = "protoc-gen-openapi"
+)
+
+func Command(ctx context.Context, args ...string) *exec.Cmd {
+	sg.Deps(ctx, PrepareCommand)
+	return sg.Command(ctx, sg.FromBinDir(name), args...)
+}
+
+func PrepareCommand(ctx context.Context) error {
+	_, err := sgtool.GoInstall(ctx, "github.com/google/gnostic/cmd/"+name, "v"+version)
```
fredrikaverpil, line 26: @alethenorio actually, @radhus mentioned something ... why don't we do a `go get` followed by `go build` and finally a symlink into .sage/bin?

Here we perform a `go install` which would mess with the binary you might already have on your local system.

radhus:
> Does the package name need to contain the openapi version? The tool seems to be generic so maybe `sgprotocgenopenapi` is enough?

Maybe even calling it `sggnostic` as the tool here is Gnostic and there are other protobuf compilers for OpenAPI 🤓

fredrikaverpil:
> Maybe even calling it sggnostic as the tool here is Gnostic and there are other protobuf compilers for OpenAPI 🤓

The binary is `protoc-gen-openapi` and you could potentially add another "sgtool" to Sage which downloads and installs a differently named binary from the gnostic project. See here:
https://github.com/google/gnostic/tree/main/cmd/

tools/sgprotocgenopenapi/tools.go
alethenorio: Why would it mess with your local binary? The GOBIN env variable guarantees that it is built and placed in the right place and not in the Go bin folder

fredrikaverpil: Oh wait, the command is only built on line 26 and not executed!
So that means that when the actual `go install` command runs, it runs with the `GOBIN` set, which means it builds into the `GOBIN` dir, which is `.sage/bin`?

alethenorio: Correct

fredrikaverpil: Nice, sorry for the noise 😅 I don't think there's a problem at all here then, or what do you say @radhus

alethenorio (approved):

PR #621, opened 2025-01-08
feat: bump downloaded go to version 1.23.4

thall:
If Sage is invoked in an environment where Go is not available in `$PATH`, Sage will download it.

This should not affect any users that has Go installed outside of an Sage repo.


radhus (approved):
Makes sense since it's only affecting the downloaded version! Need to regenerate Makefile in Sage repos though (i.e. in this PR as well?).

PR #622, opened 2025-01-10
fix: use proper binary name for v2+ modules

kristofferwanglund:
When installing a go module which has made a major version bump,
e.g. example.com/foo => example.com/foo/v2,
just looking at the last element of the path to get the binary name
returns "v2" instead of "foo" (which is what 'go install' will name
the binary).

Instead use a 'go list' helper function to determine the name.


radhus (approved):
Makes sense!

PR #636, opened 2025-02-13
fix(CODEOWNERS): remove overlapping group

radhus:
We're (almost) all part of the latter group, and for some reason that makes me get double notifications for all PRs...

I guess we could organize the groups somehow so that the former group members are always included in the latter group?

For @alethenorio to review 😄


radhus:
Not possible to organize groups that way apparently 😭

PR #638, opened 2025-02-14
Add support to run `gopls check`

joaoeinride:
[gopls](https://pkg.go.dev/golang.org/x/tools/gopls) supports running as a linter through the CLi with `gopls check <PATH_TO_GO_FILE> ...`

### DIFF
--- a/tools/sggopls/tools.go
+++ b/tools/sggopls/tools.go
@@ -1,0 +1,32 @@
+package sggopls
+
+import (
+	"context"
+	"os/exec"
+
+	"go.einride.tech/sage/sg"
+	"go.einride.tech/sage/sgtool"
+)
+
+const (
+	name    = "gopls"
+	version = "v0.17.1"
+)
+
+// Check runs `gopls check <PATH1> <PATH2>`.
+// gopls only works with paths to go files.
+func Check(ctx context.Context, paths []string) *exec.Cmd {
+	sg.Deps(ctx, PrepareCommand)
+	args := []string{"check"}
+	args = append(args, paths...)
+	return sg.Command(
+		ctx,
+		name,
+		args...,
+	)
+}
+
+func PrepareCommand(ctx context.Context) error {
+	_, err := sgtool.GoInstall(ctx, "golang.org/x/tools/gopls", version)
+	return err
+}


tools/sggopls/tools.go
```diff
@@ -0,0 +1,53 @@
...
+	"go.einride.tech/sage/sgtool"
+)
+
+const (
+	name    = "gopls"
+	version = "v0.17.1"
+)
+
+// Root will find all go files in subdirectories from the project git root.
+func Root(ctx context.Context) error {
```
alethenorio, line 19: What's the performance for such a method? There can be a lot of files in a project, has this been tested with a fairly large project?

tools/sggopls/tools.go
```diff
@@ -0,0 +1,53 @@
...
+	de, err := os.ReadDir(sg.FromGitRoot("."))
```
alethenorio, line 20:
```suggestion
	de, err := os.ReadDir(sg.FromGitRoot())
```

tools/sggopls/tools.go
```diff
@@ -0,0 +1,53 @@
...
+	if err != nil {
+		return err
+	}
+	var paths []string
```
alethenorio, line 24: I think `filepath.WalkDir`might be more performant, no?

tools/sggopls/tools.go
joaoeinride: I guess we can write this or a similar version of this for our own use case. I will remove this

tools/sggopls/tools.go
joaoeinride: will do

tools/sggopls/tools.go
joaoeinride: removed

alethenorio:
@joaoeinride I am wondering about the use case for this package. It feels like the main use case for `gopls` is to be integrated with a code editor rather than running as a CLI. Does it check for anything that `golangci-lint` doesn't already do?

Its a bit of a strange use case to have to run it per Go file. Feels like this is not quite the intended use case.

fredrikaverpil:
@joaoeinride @alethenorio I'm echoing what Alex said, but primarily because I'm curious about your use case!

joaoeinride:
> @joaoeinride I am wondering about the use case for this package. It feels like the main use case for gopls is to be integrated with a code editor rather than running as a CLI. Does it check for anything that golangci-lint doesn't already do?

gopls runs linters I haven't found in golangci-lint, see https://cs.opensource.google/go/x/tools/+/refs/tags/gopls/v0.17.1:gopls/doc/analyzers.md . For example it detected the misuse of an iterator yield function.

> Its a bit of a strange use case to have to run it per Go file. Feels like this is not quite the intended use case.

I get that, I also realized it "unstable". According to

```
 [Command-line interface](https://cs.opensource.google/go/x/tools/+/refs/tags/gopls/v0.17.1:gopls/doc/command-line.md#): CLI for debugging and scripting (unstable)
```

from https://cs.opensource.google/go/x/tools/+/refs/tags/gopls/v0.17.1:gopls/doc/features/README.md

So I get if you don't want to add this.

fredrikaverpil:
I say why not?
Let's try it out.

alethenorio (approved):
It seems for the most part they are making use of `go vet` and `staticcheck` both which we should get outside of `gopls` but they do have a few internal [analyzers](https://pkg.go.dev/golang.org/x/tools/internal/lsp/analysis).

Let's give it a try given the simple API

fredrikaverpil (approved):

fredrikaverpil:
@joaoeinride  I'm actually on vacation so I haven't checked thoroughly. But would it be worth adding the analyzers explicitly in sage?

joaoeinride:
@fredrikaverpil no problem in doing so. I would suggest in a separate PR though

PR #639, opened 2025-02-14
feat: check repository with gopls check

joaoeinride:


fredrikaverpil (approved):
Maybe include this link?
https://github.com/golang/tools/blob/master/gopls/doc/analyzers.md

fredrikaverpil:
@joaoeinride you didn't run `gopls` from sage's bin dir. I pushed a commit to fix that. Feel free to squash it.

There are some linting errors now:

```
[go-pls] Linting Go files with gopls...
[go-pls] 
/Users/fredrik/code/work/public/sage/tools/sggolicenses/tools.go:56:53-67: runtime.GOROOT is deprecated: The root used during the Go build will not be meaningful if the binary is copied to another machine. Use the system path t
o locate the “go” binary, and use “go env GOROOT” to find its GOROOT.
/Users/fredrik/code/work/public/sage/tools/sggolicenses/tools.go:100:54-68: runtime.GOROOT is deprecated: The root used during the Go build will not be meaningful if the binary is copied to another machine. Use the system path 
to locate the “go” binary, and use “go env GOROOT” to find its GOROOT.
```

PR #640, opened 2025-02-17
feat(api-linter): bump to v1.69.1

fredrikaverpil:
This adds and fixes some AIP linters.

## Warning

[v1.68.0](https://github.com/googleapis/api-linter/releases/tag/v1.68.0) says:

> update to min version to Go 1.22 (#1457) (f34f16b)

Could this be a blocker for upgrading sage on projects which are on Go < v.22 😬

### DIFF
--- a/tools/sgapilinter/tools.go
+++ b/tools/sgapilinter/tools.go
@@ -20,7 +20,7 @@ import (
 )
 
 const (
-	version = "1.67.2"
+	version = "1.69.1"
 	name    = "api-linter"
 )
 


radhus (approved):
> Could this be a blocker for upgrading sage on projects which are on Go < v.22 😬

Sage isn't using the Go code from aip-linter as it's just downloading a prebuilt binary, so I think it should be fine!

fredrikaverpil:
> Sage isn't using the Go code from aip-linter as it's just downloading a prebuilt binary, so I think it should be fine!

I meant implicitly blocking updating Sage. Or maybe more likely, having to disable AIP tests...

Anyway, should probably announce this change.

PR #369, opened 2023-01-04
fix(go-licences): fix running on macOS when Go is upgraded

radhus:
Running go-licenses breaks after upgrading Go through Homebrew, as the GOROOT path on the machine no longer matches the one used to build go-licenses.

As a workaround, explicitly set GOROOT.

See https://github.com/google/go-licenses/issues/149 for reference.

PR #641, opened 2025-02-17
fix: remove GOROOT from go-licenses

fredrikaverpil:
### Why?

In #639 I noticed a gopls warning:

```
[go-pls] Linting Go files with gopls...
[go-pls] 
/Users/fredrik/code/work/public/sage/tools/sggolicenses/tools.go:56:53-67: runtime.GOROOT is deprecated: The root used during the Go build will not be meaningful if the binary is copied to another machine. Use the system path t
o locate the “go” binary, and use “go env GOROOT” to find its GOROOT.
/Users/fredrik/code/work/public/sage/tools/sggolicenses/tools.go:100:54-68: runtime.GOROOT is deprecated: The root used during the Go build will not be meaningful if the binary is copied to another machine. Use the system path 
to locate the “go” binary, and use “go env GOROOT” to find its GOROOT.
```

### What?

I removed the line setting the GOROOT env var.

### Notes

@alethenorio I believe you originally wrote this?

https://github.com/einride/sage/blob/ea6326b1a84a8551b58b380d000c56bb37496cf0/tools/sggolicenses/tools.go#L47-L56

Please have a look :smile:

### DIFF
--- a/tools/sggolicenses/tools.go
+++ b/tools/sggolicenses/tools.go
@@ -44,16 +44,6 @@ func CheckDir(ctx context.Context, directory string, disallowedTypes ...string) 
 	}
 	cmd := Command(ctx, args...)
 	cmd.Dir = directory
-	// go-licenses tries to exclude standard library packages by checking if they are prefixed
-	// with `runtime.GOROOT()`. However, if the go-licenses tool is not run with a GOROOT environment variable,
-	// that call will return the GOROOT path used during build time of go-licenses. This typically works on Linux,
-	// but on macOS with Homebrew, the GOROOT is version prefixed, which breaks as soon as Go is upgraded.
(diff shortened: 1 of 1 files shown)


radhus:
#369 is the source of that addition, and the [go-licenses issue](https://github.com/google/go-licenses/issues/149) is still open. If you can test with e.g. a go-licenses built with Go <1.x and have Go >1.x installed then that would be nice 😄 

Or we just revert like this and let users complain if go-licenses start to error again!

alethenorio:
GOROOT was deprecated in Go 1.24 so I don't think we should remove it just yet until we migrate Sage to ir

fredrikaverpil:
I guess we can close this for now or put it into draft for later.

PR #654, opened 2025-03-10
feat: GoInstallWithPost

fredrikaverpil:
### Why?

Some tools wants us to install using a "post" version suffix. An example can be seen [here](https://vektra.github.io/mockery/latest/installation/) where `v2` is the "post":

```bash
go install github.com/vektra/mockery/v2@v2.53.1
```

But our current `GoInstall` does not support adding this in, or the symlink in the `.sage/bin` folder will be named `v2`.

### What?

Add new `GoInstallWithPost` function.


sgtool/build.go
```diff
@@ -35,6 +35,33 @@ func GoInstall(ctx context.Context, pkg, version string) (string, error) {
 	return symlink, nil
 }
 
+func GoInstallWithPost(ctx context.Context, pkg, post, version string) (string, error) {
```
radhus, line 38: What's the relation to #622? Can we reuse the same logic in both places (here and `GoinstallWithModfile`? 😇

fredrikaverpil: I tried using this instead just to see what happens:

```go
func PrepareCommand(ctx context.Context) error {
	_, err := sgtool.GoInstallWithModfile(ctx, "github.com/vektra/mockery/v2", version)
	return err
}
```

```bash
❯ make generate-mocks
[generate-mocks] generating...
[github:com/einride/freight-planner-service/:sage/mockery:prepare-command] no required module provides package github.com/vektra/mockery/v2; to add it:
[github:com/einride/freight-planner-service/:sage/mockery:prepare-command]      go get github.com/vektra/mockery/v2
[github:com/einride/freight-planner-service/:sage/mockery:prepare-command] exit status 1
make: *** [generate-mocks] Error 1
```

It seems that expects me to also add mockery to the sage's `go.mod`... 🤔 

@kristofferwanglund hey, do you have a minute to look at this together?

PR #661, opened 2025-03-24
feat(sggolangcilintv2): add golangci-lint v2

fredrikaverpil:
### Why?

Golangci-lint was bumped to v2.

### What?

- Add `sggolangcilintv2` as there are backwards braking changes to the CLI. The CLI arguments will differ from golangci-lint v1 and options around exclusion paths have been moved from the CLI so they must be defined in the config yaml.
- Deprecate `sggolangcilint`.
- Use golangci-lint v2.1.1.
- Run `golangci-lint migrate -c golangci.yml`. This sorted the linters for us, so they are now in alphabetical order. The `gci`, `gofumpt` linters were removed (now exists as formatters) and `gosimple` has been absorbed by `staticcheck`.
- Add comments back in which were removed by the migration tool.
- Address new linting issues found by golangci-lint.
- Use templating to allow per-project customization.

### Will be saved for a separate PR

- Formatting capabilities: https://github.com/einride/sage/pull/662
- Discussing adding/removing/reconfiguring [linters](https://golangci-lint.run/usage/linters/) - can be saved for backend guild session.

<details>
<summary>Enabled vs disabled linters</summary>

Name | Description | Enabled
-- | -- | -- 
asasalint | Check for pass []any as any in variadic func(...any). | ✅ 
asciicheck | Checks that all code identifiers does not have non-ASCII symbols in the name. | ✅ 
bidichk | Checks for dangerous unicode character sequences. | ✅ 
bodyclose | Checks whether HTTP response body is closed successfully. | ✅ 
canonicalheader | Canonicalheader checks whether net/http.Header uses canonical header. |  
containedctx | Containedctx is a linter that detects struct contained context.Context field. |   
contextcheck | Check whether the function uses a non-inherited context. |   
copyloopvar | A linter detects places where loop variables are copied. | ✅  
cyclop | Checks function and package cyclomatic complexity. |   
decorder | Check declaration order and count of types, constants, variables and functions. |   
depguard | Go linter that checks if package imports are in a list of acceptable packages. |   
dogsled | Checks assignments with too many blank identifiers (e.g. x, , , _, := f()). |   
dupl | Detects duplicate fragments of code. |   
dupword | Checks for duplicate words in the source code. |  
durationcheck | Check for two durations multiplied together. | ✅ 
err113 | Go linter to check the errors handling expressions. |  
errcheck | Check for unchecked errors. Unchecked errors can be critical bugs in some cases. |  ✅ 
errchkjson | Checks types passed to the json encoding functions. Reports unsupported types and reports occurrences where the check for the returned error can be omitted. | ✅
errname | Checks that sentinel errors are prefixed with the Err and error types are suffixed with the Error. | ✅ 
errorlint | Errorlint is a linter for that can be used to find code that will cause problems with the error wrapping scheme introduced in Go 1.13. | ✅ 
exhaustive | Check exhaustiveness of enum switch statements. |   
exhaustruct | Checks if all structure fields are initialized. |   
exptostd | Detects functions from golang.org/x/exp/ that can be replaced by std functions. | ✅ 
fatcontext | Detects nested contexts in loops and function literals. |  
forbidigo | Forbids identifiers. |   
forcetypeassert | Finds forced type assertions. |   
funlen | Checks for long functions. |   
ginkgolinter | Enforces standards of using ginkgo and gomega. | 
gocheckcompilerdirectives | Checks that go compiler directive comments (//go:) are valid. |   
gochecknoglobals | Check that no global variables exist. | ✅ 
gochecknoinits | Checks that no init functions are present in Go code. | ✅ 
gochecksumtype | Run exhaustiveness checks on Go "sum types". |   
gocognit | Computes and checks the cognitive complexity of functions. |   
goconst | Finds repeated strings that could be replaced by a constant. |   
gocritic | Provides diagnostics that check for bugs, performance and style issues.Extensible without recompilation through dynamic rules.Dynamic rules are written declaratively with AST patterns, filters, report message and optional suggestion. | ✅ 
gocyclo | Computes and checks the cyclomatic complexity of functions. |   
godot | Check if comments end in a period. | ✅ 
godox | Detects usage of FIXME, TODO and other keywords inside comments. |   
goheader | Checks if file header matches to pattern. |  
gomoddirectives | Manage the use of 'replace', 'retract', and 'excludes' directives in go.mod. | ✅  
gomodguard | Allow and block list linter for direct Go module dependencies. This is different from depguard where there are different block types for example version constraints and module recommendations. | ✅ 
goprintffuncname | Checks that printf-like functions are named with f at the end. | ✅ 
gosec | Inspects source code for security problems. | ✅ 
gosmopolitan | Report certain i18n/l10n anti-patterns in your Go codebase. |   
govet | Report suspicious constructs, such as Printf calls whose arguments do not align with the format string. | ✅ 
grouper | Analyze expression groups. |   
iface | Detect the incorrect use of interfaces, helping developers avoid interface pollution. |  
importas | Enforces consistent import aliases. |  
inamedparam | Reports interfaces with unnamed method parameters. |   
ineffassign | Detect when assignments to existing variables are not used. | ✅ 
interfacebloat | A linter that checks the number of methods inside an interface. |   
intrange | Intrange is a linter to find places where for loops could make use of an integer range. |  
ireturn | Accept Interfaces, Return Concrete Types. |   
lll | Reports long lines. | ✅ 
loggercheck | Checks key value pairs for common logger libraries (kitlog,klog,logr,zap). |   
maintidx | Maintidx measures the maintainability index of each function. |   
makezero | Finds slice declarations with non-zero initial length. | ✅ 
mirror | Reports wrong mirror patterns of bytes/strings usage. |  
misspell | Finds commonly misspelled English words. | ✅ 
mnd | An analyzer to detect magic numbers. |   
musttag | Enforce field tags in (un)marshaled structs. |   
nakedret | Checks that functions with naked returns are not longer than a maximum size (can be zero). | ✅ 
nestif | Reports deeply nested if statements. |   
nilerr | Finds the code that returns nil even if it checks that the error is not nil. | ✅  
nilnesserr | Reports constructs that checks for err != nil, but returns a different nil value error.Powered by nilness and nilerr. |   
nilnil | Checks that there is no simultaneous return of nil error and an invalid value. |   
nlreturn | Nlreturn checks for a new line before return and branch statements to increase code clarity. |  
noctx | Finds sending http request without context.Context. | ✅ 
nolintlint | Reports ill-formed or insufficient nolint directives. | ✅ 
nonamedreturns | Reports all named returns. |   
nosprintfhostport | Checks for misuse of Sprintf to construct a host with port in a URL. |   
paralleltest | Detects missing usage of t.Parallel() method in your Go test. |   
perfsprint | Checks that fmt.Sprintf can be replaced with a faster alternative. |  
prealloc | Finds slice declarations that could potentially be pre-allocated. | ✅ 
predeclared | Find code that shadows one of Go's predeclared identifiers. | ✅ 
promlinter | Check Prometheus metrics naming via promlint. | ✅ 
protogetter | Reports direct reads from proto message fields when getters should be used. | ✅ 
reassign | Checks that package variables are not reassigned. |   
recvcheck | Checks for receiver type consistency. |   
revive | Fast, configurable, extensible, flexible, and beautiful linter for Go. Drop-in replacement of golint. | ✅ 
rowserrcheck | Checks whether Rows.Err of rows is checked successfully. | ✅ 
sloglint | Ensure consistent code style when using log/slog. | ✅ 
spancheck | Checks for mistakes with OpenTelemetry/Census spans. | ✅ 
staticcheck | Go vet on steroids, applying a ton of static analysis checks. | ✅ 
sqlclosecheck | Checks that sql.Rows, sql.Stmt, sqlx.NamedStmt, pgx.Query are closed. | ✅ 
tagalign | Check that struct tags are well aligned. |  
tagliatelle | Checks the struct tags. |   
testableexamples | Linter checks if examples are testable (have an expected output). |   
testifylint | Checks usage of github.com/stretchr/testify. |  
testpackage | Linter that makes you use a separate _test package. |   
thelper | Thelper detects tests helpers which is not start with t.Helper() method. |   
tparallel | Tparallel detects inappropriate usage of t.Parallel() method in your Go test codes. | ✅ 
unconvert | Remove unnecessary type conversions. | ✅ 
unused | Check Go code for unused constants, variables, functions and types | ✅ 
unparam | Reports unused function parameters. |   
usestdlibvars | A linter that detect the possibility to use variables/constants from the Go standard library. | ✅ 
usetesting | Reports uses of functions with replacement inside the testing package. |  
varnamelen | Checks that the length of a variable's name matches its scope. |   
wastedassign | Finds wasted assignment statements. | ✅ 
whitespace | Whitespace is a linter that checks for unnecessary newlines at the start and end of functions, if, for, etc. | ✅ 
wrapcheck | Checks that errors returned from external packages are wrapped. |   
wsl | Add or remove empty lines. |
zerologlint | Detects the wrong usage of zerolog that a user forgets to dispatch with Send or Msg. |  

</details>

### Notes

- https://golangci-lint.run/product/migration-guide/
- https://golangci-lint.run/usage/formatters/

### DIFF
--- a/internal/codegen/imports.go
+++ b/internal/codegen/imports.go
@@ -99,6 +99,7 @@ func importPathToAssumedName(importPath string) string {
 	}
 	base = strings.TrimPrefix(base, "go-")
 	if i := strings.IndexFunc(base, func(r rune) bool {
+		//nolint:staticcheck // De Morgan's law intentionally not applied
 		return !('a' <= r && r <= 'z' || 'A' <= r && r <= 'Z' || '0' <= r && r <= '9' || r == '_' ||
 			r >= utf8.RuneSelf && (unicode.IsLetter(r) || unicode.IsDigit(r)))
 	}); i >= 0 {
--- a/.sage/main.go
+++ b/.sage/main.go
@@ -12,7 +12,7 @@ import (
 	"go.einride.tech/sage/tools/sgconvco"
 	"go.einride.tech/sage/tools/sggit"
 	"go.einride.tech/sage/tools/sggo"
-	"go.einride.tech/sage/tools/sggolangcilint"
+	"go.einride.tech/sage/tools/sggolangcilintv2"
 	"go.einride.tech/sage/tools/sggolicenses"
 	"go.einride.tech/sage/tools/sggolines"
 	"go.einride.tech/sage/tools/sggopls"
@@ -48,7 +48,10 @@ func GoTest(ctx context.Context) error {
 
 func GoLint(ctx context.Context) error {
 	sg.Logger(ctx).Println("linting Go files...")
-	return sggolangcilint.Run(ctx)
+	return sggolangcilintv2.Run(
+		ctx,
+		sggolangcilintv2.Config{RunRelativePathMode: sggolangcilintv2.RunRelativePathModeGitRoot},
+	)
 }
 
 func GoPls(ctx context.Context) error {
@@ -62,7 +65,10 @@ func GoPls(ctx context.Context) error {
 
 func GoLintFix(ctx context.Context) error {
 	sg.Logger(ctx).Println("fixing Go files...")
-	return sggolangcilint.Fix(ctx)
+	return sggolangcilintv2.Fix(
+		ctx,
+		sggolangcilintv2.Config{RunRelativePathMode: sggolangcilintv2.RunRelativePathModeGitRoot},
+	)
 }
 
 func GoFormat(ctx context.Context) error {
--- a/tools/sgartifactregistry/auth.go
+++ b/tools/sgartifactregistry/auth.go
@@ -57,9 +57,9 @@ func NpmAuthenticate(ctx context.Context, packageJSONDir, registryURL string) er
 		yarnMajor = strings.Split(version, ".")[0]
 	}
 
-	switch {
+	switch yarnMajor {
 	// If yarn v1 or npm we use npm config
-	case yarnMajor == "1":
+	case "1":
 		cmd = sg.Command(
 			ctx,
 			"npm",
--- a/tools/sggolangcilintv2/golangci.yml.tmpl
+++ b/tools/sggolangcilintv2/golangci.yml.tmpl
@@ -1,0 +1,263 @@
+version: "2"
+
+run:
+  build-tags:
+    - wireinject
+  relative-path-mode: {{ .RunRelativePathMode }}
+
+issues:
+  fix: false
+
+linters:
+  default: none
+  enable:
+    # Check for pass []any as any in variadic func(...any).
+    # [fast: false, auto-fix: false]
+    - asasalint
+
+    # Check that code does not contain non-ASCII identifiers.
+    # [fast: true, auto-fix: false]
+    - asciicheck
+
+    # Checks for dangerous unicode character sequences.
+    # [fast: true]
+    - bidichk
+
+    # Check whether HTTP response body is closed successfully.
+    # [fast: false, auto-fix: false]
+    - bodyclose
+
+    # Copyloopvar is a linter detects places where loop variables are copied.
+    # [fast: false, auto-fix: false]
+    - copyloopvar
+
+    # Check for two durations multiplied together.
+    # [fast: false, auto-fix: false]
+    - durationcheck
+
+    # Check for unchecked errors. Unchecked errors can be critical bugs in some cases.
+    # [fast: false, auto-fix: false]
+    - errcheck
+
+    # Checks types passed to the json encoding functions. Reports unsupported types and reports occurrences where the check for the returned error can be omitted.
+    # [fast: false, autofix: false]
+    - errchkjson
+
+    # Check that sentinel errors are prefixed with the `Err` and error types are suffixed with the `Error`.
+    # [fast: false, auto-fix: false]
+    - errname
+
+    # Find code that will cause problems with the error wrapping scheme introduced in Go 1.13.
+    # [fast: false, auto-fix: false]
+    - errorlint
+
+    # Detects functions from golang.org/x/exp/ that can be replaced by std functions.
+    # [fast: ?, auto-fix: true]
+    - exptostd
+
+    # Check that no global variables exist.
+    # [fast: false, auto-fix: false]
+    - gochecknoglobals
+
+    # Check that no init functions are present in Go code.
+    # [fast: true, auto-fix: false]
+    - gochecknoinits
+
+    # Check for bugs, performance and style issues.
+    # [fast: false, auto-fix: false]
+    - gocritic
+
+    # Check if comments end in a period.
+    # [fast: true, auto-fix: true]
+    - godot
+
+    # Manage the use of 'replace', 'retract', and 'excludes' directives in go.mod.
+    # [fast: true, auto-fix: false]
+    - gomoddirectives
+
+    # Allow and block list linter for direct Go module dependencies. This is different from depguard where there are different block types for example version constraints and module recommendations.
+    # [fast: true, auto-fix: false]
+    - gomodguard
+
+    # Check that printf-like functions are named with `f` at the end.
+    # [fast: true, auto-fix: false]
+    - goprintffuncname
+
+    # Inspects source code for security problems.
+    # [fast: false, auto-fix: false]
+    - gosec
+
+    # Report suspicious constructs, such as Printf calls whose arguments do not align with the format string.
+    # [fast: false, auto-fix: false]
+    - govet
+
+    # Detect when assignments to existing variables are not used.
+    # [fast: true, auto-fix: false]
+    - ineffassign
+
+    # Report long lines.
+    # [fast: true, auto-fix: false]
+    - lll
+
+    # Find slice declarations with non-zero initial length.
+    # [fast: false, auto-fix: false]
+    - makezero
+
+    # Find commonly misspelled English words in comments.
+    # [fast: true, auto-fix: true]
+    - misspell
+
+    # Find naked returns in functions greater than a specified function length.
+    # [fast: true, auto-fix: false]
+    - nakedret
+
+    # Find the code that returns nil even if it checks that the error is not nil.
+    # [fast: false, auto-fix: false]
+    - nilerr
+
+    # Finds sending http request without context.Context.
+    # [fast: false, auto-fix: false]
+    - noctx
+
+    # Report ill-formed or insufficient nolint directives.
+    # [fast: true, auto-fix: false]
+    - nolintlint
+
+    # Find slice declarations that could potentially be preallocated.
+    # [fast: true, auto-fix: false]
+    - prealloc
+
+    # Find code that shadows one of Go's predeclared identifiers.
+    # [fast: true, auto-fix: false]
+    - predeclared
+
+    # Check Prometheus metrics naming via promlint.
+    # [fast: true, auto-fix: false]
+    - promlinter
+
+    # Reports direct reads from proto message fields when getters should be used.
+    # [fast: false, auto-fix: true]
+    - protogetter
+
+    # Fast, configurable, extensible, flexible, and beautiful linter for Go. Drop-in replacement of golint.
+    # [fast: false, auto-fix: true]
+    - revive
+
+    # Check whether Err of rows is checked successfully.
+    # [fast: false, auto-fix: false]
+    - rowserrcheck
+
+    # Ensure consistent code style when using log/slog.
+    # [fast: false, auto-fix: false]
+    - sloglint
+
+    # Checks for mistakes with OpenTelemetry/Census spans.
+    # [fast: false, auto-fix: false]
+    - spancheck
+
+    # Check that sql.Rows and sql.Stmt are closed.
+    # [fast: false, auto-fix: false]
+    - sqlclosecheck
+
+    # Go vet on steroids, applying a ton of static analysis checks.
+    # [fast: false, auto-fix: false]
+    - staticcheck
+
+    # Detect inappropriate usage of t.Parallel() method in your Go test codes.
+    # [fast: false, auto-fix: false]
+    - tparallel
+
+    # Remove unnecessary type conversions.
+    # [fast: false, auto-fix: false]
+    - unconvert
+
+    # Check Go code for unused constants, variables, functions and types.
+    # [fast: false, auto-fix: false]
+    - unused
+
+    # A linter that detect the possibility to use variables/constants from the Go standard library.
+    # [fast: true, auto-fix: true]
+    - usestdlibvars
+
+    # Find wasted assignment statements.
+    # [fast: false, auto-fix: false]
+    - wastedassign
+
+    # Detect leading and trailing whitespace.
+    # [fast: true, auto-fix: true]
+    - whitespace
+  settings:
+    errorlint:
+      errorf: false
+
+    gomoddirectives:
+      replace-local: true
+
+    gomodguard:
+      blocked:
+        modules:
+          - github.com/sirupsen/logrus:
+              recommendations:
+                - go.uber.org/zap
+              reason: Use Zap for logging.
+          - github.com/rs/zerolog:
+              recommendations:
+                - go.uber.org/zap
+              reason: Use Zap for logging.
+          - github.com/pkg/errors:
+              recommendations:
+                - errors
+                - fmt
+              reason: Use the standard library error packages.
+          - golang.org/x/xerrors:
+              recommendations:
+                - errors
+                - fmt
+              reason: Use the standard library error packages.
+          - github.com/golang/protobuf:
+              recommendations:
+                - google.golang.org/protobuf
+              reason: The protobuf v1 module is deprecated.
+          - github.com/stretchr/testify:
+              recommendations:
+                - gotest.tools/v3
+              reason: More reliably supports protobuf messages.
+
+    gosec:
+      excludes:
+        - G115
+
+    lll:
+      line-length: 120
+      tab-width: 1
+
+    misspell:
+      locale: US
+      ignore-rules:
+        - cancelled
+        - cancelling
+        - analyses
+
+    prealloc:
+      simple: true
+      range-loops: true
+      for-loops: true
+
+    protogetter:
+      skip-any-generated: true
+
+    staticcheck:
+      checks:
+        - all
+
+  exclusions:
+    generated: lax
+    presets:
+      - comments
+      - common-false-positives
+      - legacy
+      - std-error-handling
+    {{- if .LintersExclusionsPaths }}
+    paths:{{ range .LintersExclusionsPaths }}
+      - {{ . }}{{ end }}
+    {{- end }}
--- a/tools/sggolangcilintv2/tools.go
+++ b/tools/sggolangcilintv2/tools.go
@@ -1,0 +1,173 @@
+// This package is considered experimental and may change.
+
+package sggolangcilintv2
+
+import (
+	"context"
+	_ "embed"
+	"errors"
+	"fmt"
+	"io/fs"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"text/template"
+
+	"go.einride.tech/sage/sg"
+	"go.einride.tech/sage/sgtool"
+)
+
+const (
+	name    = "golangci-lint"
+	version = "2.1.1"
+
+	RunRelativePathModeGitRoot    = "gitroot"
+	RunRelativePathModeGomod      = "gomod"
+	RunRelativePathModeCfg        = "cfg" // WARNING: not recommended; paths will be relative to .sage/tools/golangci-lint
+	RunRelativePathModeWorkingDir = "wd"  // WARNING: not recommended according to official docs
+)
+
+type RunRelativePathMode string
+
+//go:embed golangci.yml.tmpl
+var configTemplate []byte
+
+// Config holds values which will be applied to the template, ultimately resulting in the final golangci.yml file.
+type Config struct {
+	// The mode used to evaluate relative paths. It's used by exclusions, Go plugins and some linters.
+	RunRelativePathMode RunRelativePathMode
+	// Which file paths to exclude from issue reporting. The paths will still be analyzed.
+	// Golangci-Lint will:
+	// - replace "/" with the current OS file path separator to properly work on Windows.
+	// - allow you to optionally use regexps here, like ".*\\.my\\.go$".
+	// - treat paths relative to the setting of RunRelativePathMode.
+	LintersExclusionsPaths []string
+}
+
+func Command(ctx context.Context, config Config, args ...string) *exec.Cmd {
+	sg.Deps(ctx, func(ctx context.Context) error {
+		return PrepareCommand(ctx, config)
+	})
+	return sg.Command(ctx, sg.FromBinDir(name), args...)
+}
+
+func defaultConfigPath() string {
+	return sg.FromToolsDir(name, ".golangci.yml")
+}
+
+func CommandInDirectory(ctx context.Context, config Config, directory string, args ...string) *exec.Cmd {
+	configPath := filepath.Join(directory, ".golangci.yml")
+	if _, err := os.Lstat(configPath); errors.Is(err, os.ErrNotExist) {
+		configPath = defaultConfigPath()
+	}
+	cmdArgs := append([]string{"run", "--allow-parallel-runners", "-c", configPath}, args...)
+	cmd := Command(ctx, config, cmdArgs...)
+	cmd.Dir = directory
+	return cmd
+}
+
+// Run GolangCI-Lint in every Go module from the root of the current git repo.
+func Run(ctx context.Context, config Config, args ...string) error {
+	var commands []*exec.Cmd
+	if err := filepath.WalkDir(sg.FromGitRoot(), func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if d.IsDir() || d.Name() != "go.mod" {
+			return nil
+		}
+		cmd := CommandInDirectory(ctx, config, filepath.Dir(path), args...)
+		commands = append(commands, cmd)
+		return cmd.Start()
+	}); err != nil {
+		return err
+	}
+	errs := make([]error, 0, len(commands))
+	for _, cmd := range commands {
+		errs = append(errs, cmd.Wait())
+	}
+	return errors.Join(errs...)
+}
+
+// Run GolangCI-Lint --fix in every Go module from the root of the current git repo.
+func Fix(ctx context.Context, config Config, args ...string) error {
+	var commands []*exec.Cmd
+	if err := filepath.WalkDir(sg.FromGitRoot(), func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if d.IsDir() || d.Name() != "go.mod" {
+			return nil
+		}
+		cmd := Command(
+			ctx,
+			config,
+			append([]string{"run", "--allow-serial-runners", "-c", defaultConfigPath(), "--fix"}, args...)...,
+		)
+		cmd.Dir = filepath.Dir(path)
+		commands = append(commands, cmd)
+		return cmd.Start()
+	}); err != nil {
+		return err
+	}
+	for _, cmd := range commands {
+		if err := cmd.Wait(); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func PrepareCommand(ctx context.Context, config Config) error {
+	toolDir := sg.FromToolsDir(name)
+	binDir := filepath.Join(toolDir, version, "bin")
+	binary := filepath.Join(binDir, name)
+	hostOS := runtime.GOOS
+	hostArch := runtime.GOARCH
+	golangciLint := fmt.Sprintf("golangci-lint-%s-%s-%s", version, hostOS, hostArch)
+	binURL := fmt.Sprintf(
+		"https://github.com/golangci/golangci-lint/releases/download/v%s/%s.tar.gz",
+		version,
+		golangciLint,
+	)
+	if err := sgtool.FromRemote(
+		ctx,
+		binURL,
+		sgtool.WithDestinationDir(binDir),
+		sgtool.WithUntarGz(),
+		sgtool.WithRenameFile(fmt.Sprintf("%s/golangci-lint", golangciLint), name),
+		sgtool.WithSkipIfFileExists(binary),
+		sgtool.WithSymlink(binary),
+	); err != nil {
+		return fmt.Errorf("unable to download %s: %w", name, err)
+	}
+	configPath := defaultConfigPath()
+	err := CreateConfigFromTemplate(ctx, configPath, config)
+	if err != nil {
+		return fmt.Errorf("unable to create config file: %w", err)
+	}
+
+	return nil
+}
+
+func CreateConfigFromTemplate(ctx context.Context, outputPath string, config Config) error {
+	if config.RunRelativePathMode == "" {
+		config.RunRelativePathMode = RunRelativePathModeGitRoot
+		sg.Logger(ctx).Printf("Using default relative path mode: %s\n", config.RunRelativePathMode)
+	}
+	tmpl, err := template.New("golangci.yml").Parse(string(configTemplate))
+	if err != nil {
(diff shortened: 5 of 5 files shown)


internal/codegen/imports.go
```diff
@@ -99,8 +99,8 @@ func importPathToAssumedName(importPath string) string {
 	}
 	base = strings.TrimPrefix(base, "go-")
 	if i := strings.IndexFunc(base, func(r rune) bool {
-		return !('a' <= r && r <= 'z' || 'A' <= r && r <= 'Z' || '0' <= r && r <= '9' || r == '_' ||
-			r >= utf8.RuneSelf && (unicode.IsLetter(r) || unicode.IsDigit(r)))
+		return (r < 'a' || r > 'z') && (r < 'A' || r > 'Z') && (r < '0' || r > '9') && r != '_' &&
+			(r < utf8.RuneSelf || (!unicode.IsLetter(r) && !unicode.IsDigit(r)))
```
fredrikaverpil, line 103: Looking at the error in `imports.go`, the issue is in the `ImportFunc` function where De Morgan's law can be applied to simplify a complex negation.

The original code was checking if a character is NOT a valid identifier character by negating a complex OR expression. By applying De Morgan's law:
- `!(a || b)` becomes `!a && !b`
- `!(a && b)` becomes `!a || !b`

The new version directly checks for invalid characters without the negation operator, making the code more readable and maintainable.

This transformation ensures the function's logic remains the same while satisfying the linter rule about applying De Morgan's law to simplify boolean logic.

tools/sggolangcilint/tools.go
```diff
@@ -37,12 +37,8 @@ func CommandInDirectory(ctx context.Context, directory string, args ...string) *
 	if _, err := os.Lstat(configPath); errors.Is(err, os.ErrNotExist) {
 		configPath = defaultConfigPath()
 	}
-	var excludeArg []string
-	if directory == sg.FromSageDir() {
-		excludeArg = append(excludeArg, "--exclude", "(is a global variable|is unused)")
-	}
 	cmdArgs := append([]string{"run", "--allow-parallel-runners", "-c", configPath}, args...)
-	cmd := Command(ctx, append(cmdArgs, excludeArg...)...)
+	cmd := Command(ctx, cmdArgs...)
```
fredrikaverpil, line 41: The old version:

```sh
$ golangci-lint run --help | grep exclude
  -e, --exclude strings                Exclude issue by regexp
      --exclude-use-default            Use or not use default excludes:
      --exclude-case-sensitive         If set to true exclude and exclude rules regular expressions are case-sensitive
      --exclude-files strings          Regexps of files to exclude
      --exclude-dirs strings           Regexps of directories to exclude
      --exclude-dirs-use-default       Use or not use default excluded directories:
      --exclude-generated string       Mode of the generated files analysis (default "lax")
```

The new v2.0.0 version:

```sh
❯ golangci-lint run --help
Run the linters

Usage:
  golangci-lint run [flags]

Flags:
  -c, --config PATH                       Read config from file path PATH
      --no-config                         Don't read config file
      --default string                    Default set of linters to enable (default "standard")
  -D, --disable strings                   Disable specific linter
  -E, --enable strings                    Enable specific linter
      --enable-only strings               Override linters configuration section to only run the specific linter(s)
      --fast-only                         Filter enabled linters to run only fast linters
  -j, --concurrency int                   Number of CPUs to use (Default: Automatically set to match Linux container CPU quota and fall back to the number of logical CPUs in the machine)
      --modules-download-mode string      Modules download mode. If not empty, passed as -mod=<mode> to go tools
      --issues-exit-code int              Exit code when issues were found (default 1)
      --build-tags strings                Build tags
      --timeout duration                  Timeout for total work. Disabled by default
      --tests                             Analyze tests (*_test.go) (default true)
      --allow-parallel-runners            Allow multiple parallel golangci-lint instances running.
                                          If false (default) - golangci-lint acquires file lock on start.
      --allow-serial-runners              Allow multiple golangci-lint instances running, but serialize them around a lock.
                                          If false (default) - golangci-lint exits with an error if it fails to acquire file lock on start.
      --path-prefix string                Path prefix to add to output
      --show-stats                        Show statistics per linter (default true)
      --output.text.path stdout           Output path can be either stdout, `stderr` or path to the file to write to.
      --output.text.print-linter-name     Print linter name in the end of issue text. (default true)
      --output.text.print-issued-lines    Print lines of code with issue. (default true)
      --output.text.colors                Use colors. (default true)
      --output.json.path stdout           Output path can be either stdout, `stderr` or path to the file to write to.
      --output.tab.path stdout            Output path can be either stdout, `stderr` or path to the file to write to.
      --output.tab.print-linter-name      Print linter name in the end of issue text. (default true)
      --output.tab.colors                 Use colors. (default true)
      --output.html.path stdout           Output path can be either stdout, `stderr` or path to the file to write to.
      --output.checkstyle.path stdout     Output path can be either stdout, `stderr` or path to the file to write to.
      --output.code-climate.path stdout   Output path can be either stdout, `stderr` or path to the file to write to.
      --output.junit-xml.path stdout      Output path can be either stdout, `stderr` or path to the file to write to.
      --output.junit-xml.extended         Support extra JUnit XML fields.
      --output.teamcity.path stdout       Output path can be either stdout, `stderr` or path to the file to write to.
      --output.sarif.path stdout          Output path can be either stdout, `stderr` or path to the file to write to.
      --max-issues-per-linter int         Maximum issues count per one linter. Set to 0 to disable (default 50)
      --max-same-issues int               Maximum count of issues with the same text. Set to 0 to disable (default 3)
      --uniq-by-line                      Make issues output unique by line (default true)
  -n, --new                               Show only new issues: if there are unstaged changes or untracked files, only those changes are analyzed, else only changes in HEAD~ are analyzed.
                                          It's a super-useful option for integration of golangci-lint into existing large codebase.
                                          It's not practical to fix all existing issues at the moment of integration: much better to not allow issues in new code.
                                          For CI setups, prefer --new-from-rev=HEAD~, as --new can skip linting the current patch if any scripts generate unstaged files before golangci-lint runs.
      --new-from-rev REV                  Show only new issues created after git revision REV
      --new-from-patch PATH               Show only new issues created in git patch with file path PATH
      --new-from-merge-base string        Show only new issues created after the best common ancestor (merge-base against HEAD)
      --whole-files                       Show issues in any part of update files (requires new-from-rev or new-from-patch)
      --fix                               Fix found issues (if it's supported by the linter)
      --cpu-profile-path string           Path to CPU profile output file
      --mem-profile-path string           Path to memory profile output file
      --print-resources-usage             Print avg and max memory usage of golangci-lint and total time
      --trace-path string                 Path to trace output file

Global Flags:
      --color string   Use color when printing; can be 'always', 'auto', or 'never' (default "auto")
  -h, --help           Help for a command
  -v, --verbose        Verbose output
```

internal/codegen/imports.go
```diff
@@ -99,6 +99,7 @@ func importPathToAssumedName(importPath string) string {
 	}
 	base = strings.TrimPrefix(base, "go-")
 	if i := strings.IndexFunc(base, func(r rune) bool {
+		//nolint:QF1001 // De Morgan's law intentionally not applied
 		return !('a' <= r && r <= 'z' || 'A' <= r && r <= 'Z' || '0' <= r && r <= '9' || r == '_' ||
 			r >= utf8.RuneSelf && (unicode.IsLetter(r) || unicode.IsDigit(r)))
```
fredrikaverpil, line 104: I wonder why this ignore-comment won't work... 🤔 
I feel somewhat reluctant to changing the expression although it can be auto-fixed.

Screenshot 2025-03-24 at 14 46 25

internal/codegen/imports.go
```diff
@@ -99,8 +99,8 @@ func importPathToAssumedName(importPath string) string {
 	}
 	base = strings.TrimPrefix(base, "go-")
 	if i := strings.IndexFunc(base, func(r rune) bool {
-		return !('a' <= r && r <= 'z' || 'A' <= r && r <= 'Z' || '0' <= r && r <= '9' || r == '_' ||
-			r >= utf8.RuneSelf && (unicode.IsLetter(r) || unicode.IsDigit(r)))
+		return ('a' > r || r > 'z') && ('A' > r || r > 'Z') && ('0' > r || r > '9') && r != '_' &&
+			(r < utf8.RuneSelf || !unicode.IsLetter(r) && !unicode.IsDigit(r))
```
fredrikaverpil, line 103: Please see https://github.com/einride/sage/pull/661#discussion_r2010248592

tools/sggolangcilint/golangci.yml
```diff
@@ -267,11 +267,15 @@ issues:
...
     - gci
+    - goimports
     - gofumpt
+    - golines
 
   settings:
     gofumpt:
       extra-rules: true
+    golines:
+      max-len: 120
```
fredrikaverpil, line 278: https://golangci-lint.run/usage/formatters/

tools/sggolangcilint/tools.go
```diff
@@ -91,6 +91,32 @@ func Fix(ctx context.Context, args ...string) error {
...
+func Format(ctx context.Context, args ...string) error {
+	var commands []*exec.Cmd
+	if err := filepath.WalkDir(sg.FromGitRoot(), func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if d.IsDir() || d.Name() != "go.mod" {
+			return nil
+		}
+		cmd := Command(ctx, append([]string{"fmt", "-c", defaultConfigPath()}, args...)...)
```
fredrikaverpil, line 105: Please note I did not include the `--diff` argument.

```sh
❯ golangci-lint fmt -h
Format Go source files

Usage:
  golangci-lint fmt [flags]

Flags:
  -c, --config PATH      Read config from file path PATH
      --no-config        Don't read config file
  -E, --enable strings   Enable specific formatter
  -d, --diff             Display diffs instead of rewriting files
      --stdin            Use standard input for piping source files

Global Flags:
      --color string   Use color when printing; can be 'always', 'auto', or 'never' (default "auto")
  -h, --help           Help for a command
  -v, --verbose        Verbose output
```

```sh
❯ golangci-lint help formatters
Disabled by default formatters:
gci: Checks if code and import statements are formatted, with additional rules.
gofmt: Checks if the code is formatted according to 'gofmt' command.
gofumpt: Checks if code and import statements are formatted, with additional rules.
goimports: Checks if the code and import statements are formatted according to the 'goimports' command.
golines: Checks if code is formatted, and fixes long lines.
```

fredrikaverpil: [This](https://github.com/einride/sage/actions/runs/14039963054/job/39307382299?pr=661) is what it looked like when formatting failed 🎉

Easily fixed by running:

```sh
golines --max-len=120 --write-output .
```

internal/codegen/imports.go
fredrikaverpil: Ah, `//nolint:staticcheck` worked.

fredrikaverpil:
Before merging, we should squash together 4 first commits.

tools/sggolangcilint/golangci.yml
fredrikaverpil, line 1: I propose we get rid of `copyloopvar`.

```sh
$ make go-lint

[go-lint] level=warning msg="[linters_context] copyloopvar: this linter is disabled because the Go version (1.20) of your project is lower than Go 1.22"
```

tools/sggolangcilint/golangci.yml
```diff
@@ -160,120 +171,109 @@ linters:
...
-    # https://github.com/dwmkerr/hacker-laws#hyrums-law-the-law-of-implicit-interfaces
-    errorf: false
-
-  gomoddirectives:
-    replace-local: true
+issues:
+  fix: false
+  exclude-rules:
+    - path: .sage
+      text: "(is a global variable|is unused)"
```
fredrikaverpil, line 267: I'm wondering a bit about this one.

- They removed the `--exclude` CLI argument and now we can define them here in the config file instead.
- All paths are now, by default, treated as relative to the config file.

We can configure how paths are to be treated:

```yml
# Options for analysis running.
run:
  # The mode used to evaluate relative paths.
  # It's used by exclusions, Go plugins, and some linters.
  # The value can be:
  # - `gomod`: the paths will be relative to the directory of the `go.mod` file.
  # - `gitroot`: the paths will be relative to the git root (the parent directory of `.git`).
  # - `cfg`: the paths will be relative to the configuration file.
  # - `wd` (NOT recommended): the paths will be relative to the place where golangci-lint is run.
  # Default: cfg
  relative-path-mode: gomod
```

Right now, I'm leaning towards we remove this `exclude-rule` and actually look at where this starts complaining too much. And take a decision on where/how global/unused warnings should be muted or not.

tools/sggolangcilint/golangci.yml
```diff
@@ -1,21 +1,25 @@
...
-  timeout: 10m
   build-tags:
     - wireinject
-
-issues:
-  fix: false
-  exclude-dirs:
-    - .git
-    - node_modules
+  relative-path-mode: gomod
```
fredrikaverpil, line 6: The new default is to have the linter treat all paths in the yaml as if they were relative to the config file. This makes sense if you store the `golangci.yml` in the root of your project, which we don't do in our projects.

```yml
# Options for analysis running.
run:
  # The mode used to evaluate relative paths.
  # It's used by exclusions, Go plugins, and some linters.
  # The value can be:
  # - `gomod`: the paths will be relative to the directory of the `go.mod` file.
  # - `gitroot`: the paths will be relative to the git root (the parent directory of `.git`).
  # - `cfg`: the paths will be relative to the configuration file.
  # - `wd` (NOT recommended): the paths will be relative to the place where golangci-lint is run.
  # Default: cfg
  relative-path-mode: cfg
```

So, if we want to e.g. ignore the `.git` or `node_modules` folder, we need to set this to something else. I think setting it to `gomod` might make the most sense 🤔

tools/sggolangcilint/golangci.yml
```diff
@@ -160,120 +172,109 @@ linters:
...
-    # https://github.com/dwmkerr/hacker-laws#hyrums-law-the-law-of-implicit-interfaces
-    errorf: false
-
-  gomoddirectives:
-    replace-local: true
+issues:
+  fix: false
+  exclude-rules:
+    - path: .sage
+      text: "(is a global variable|is unused)"
```
fredrikaverpil, line 268: The `--exclude` CLI argument was removed and we have to specify it here instead.

thall:
thought: Can we first upgrade golangci to v2 with as few changes as possible, so it behaves as close as today? And then later introducing formatting, so we can have a separate discussion around that?

tools/sggolangcilint/golangci.yml
```diff
@@ -160,120 +175,113 @@ linters:
...
+      max-len: 120
+
+  exclusions:
+    generated: lax
+    paths:
+      - .git
+      - node_modules
+      - third_party$
+      - builtin$
+      - examples$
```
fredrikaverpil, line 287: I'm noticing that `golangci-lint fmt` wants to format files much more heavily if this config resides inside the `.golangci.yml` (I suspect it uses default `max-len: 100`). If the formatters config is broken out into its own yaml file, it seems to behave more consistently with `max-len: 120`.

fredrikaverpil:
> thought: Can we first upgrade golangci to v2 with as few changes as possible, so it behaves as close as today? And then later introduce formatting, so we can have a separate discussion around that?

@thall Yes, I think that's good - I'll remove anything formatting related from the PR.

But just so it is clear, the formatting does not happen on `golangci-lint run`, so you won't be getting any unexpected formatting all of a sudden.

thall:
> @thall Yes, I think that's good - I'll remove anything formatting related from the PR.
> 
> But just so it is clear, the formatting does not happen on `golangci-lint run`, so you won't be getting any unexpected formatting all of a sudden.

Ah okej, that is good! Lets keep this PR as small as possible then.

tools/sggolangcilint/golangci.yml
fredrikaverpil: I actually removed this, as I don't think we need it.

fredrikaverpil:
@thall I left the linters in, which the migration tool added. It's unclear to me why it decided to add those in. I can make a separate PR where those gets added in and we can decide on them separately. Ok?

But apart from that, I think primarily we need to get the `relative-path-mode` right. I set it to `gomod` now and it seems like it everything works as expected. I can get lint warnings from the "usual suspects folders".

thall:
> @thall ~I left the linters in, which the migration tool added. It's unclear to me why it decided to add those in. I can make a separate PR where those gets added in and we can decide on them separately. Ok?~ EDIT: I just realized the "added linters" were just alphabetically sorted. So I think this PR is as constrained in scope as we can get it.
> 
> But apart from that, I think primarily we need to get the `relative-path-mode` right. I set it to `gomod` now and it seems like it everything works as expected. I can get lint warnings from the "usual suspects folders".

Perfect, looks like it just got sorted. I can try this PR out in some repos as well.

tools/sggolangcilint/golangci.yml
```diff
@@ -60,10 +71,6 @@ linters:
     # [fast: true, auto-fix: true]
     - godot
 
-    # Check whether code was gofumpt-ed.
-    # [fast: true, auto-fix: true]
-    - gofumpt
```
thall, line 65: why is this removed in the new config?

tools/sggolangcilint/golangci.yml
```diff
@@ -76,9 +83,9 @@ linters:
     # [fast: true, auto-fix: false]
     - goprintffuncname
 
-    # Check for code that can be simplified.
+    # Inspects source code for security problems.
     # [fast: false, auto-fix: false]
-    - gosimple
```
thall, line 81: why is this removed in the new config?

tools/sggolangcilint/golangci.yml
```diff
@@ -44,9 +55,9 @@ linters:
     # [fast: ?, auto-fix: true]
     - exptostd
 
-    # Check package import order and make it always deterministic.
-    # [fast: true, auto-fix: true]
-    - gci
```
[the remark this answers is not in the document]
fredrikaverpil: Removed as it was moved to formatters section.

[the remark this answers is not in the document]
fredrikaverpil: I mean, it was removed/moved by the migration tool. I can add it back in if you like.

tools/sggolangcilint/golangci.yml
fredrikaverpil: It was moved out of the `linters` section and is now a formatter, like with `gci`.

tools/sggolangcilint/golangci.yml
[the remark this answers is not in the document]
thall: Ah, looks like those are moved under formatters.
image

tools/sggolangcilint/golangci.yml
thall: Should be moved.
image

thall:
@fredrikaverpil Tried this out in one of our repos, what is the intended migration path for code like this:
```go
	return sggolangcilint.Run(
		ctx,
		"--exclude-dirs",
		"website",
		"--exclude-dirs",
		"terraform",
		"--exclude-dirs",
		"api",
		"--exclude-dirs",
		"data-mesh",
	)
```
Fails with:
```
$ make go-lint
[go-lint] linting Go files...
[go-lint] Error: unknown flag: --linters-exclusions-paths
[go-lint] Failed executing command with error: unknown flag: --linters-exclusions-paths
[go-lint] Error: unknown flag: --linters-exclusions-paths
[go-lint] Failed executing command with error: unknown flag: --linters-exclusions-paths
[go-lint] Error: unknown flag: --linters-exclusions-paths
[go-lint] Failed executing command with error: unknown flag: --linters-exclusions-paths
[go-lint] exit status 3
```

fredrikaverpil:
> what is the intended migration path for code like this

@thall unfortunately, I think exclusions now need to exist in the config file. They removed the ability to specify them as CLI args completely.

```sh
❯ golangci-lint run --help
Run the linters

Usage:
  golangci-lint run [flags]

Flags:
  -c, --config PATH                       Read config from file path PATH
      --no-config                         Don't read config file
      --default string                    Default set of linters to enable (default "standard")
  -D, --disable strings                   Disable specific linter
  -E, --enable strings                    Enable specific linter
      --enable-only strings               Override linters configuration section to only run the specific linter(s)
      --fast-only                         Filter enabled linters to run only fast linters
  -j, --concurrency int                   Number of CPUs to use (Default: Automatically set to match Linux container CPU quota and fall back to the number of logical CPUs in the machine)
      --modules-download-mode string      Modules download mode. If not empty, passed as -mod=<mode> to go tools
      --issues-exit-code int              Exit code when issues were found (default 1)
      --build-tags strings                Build tags
      --timeout duration                  Timeout for total work. Disabled by default
      --tests                             Analyze tests (*_test.go) (default true)
      --allow-parallel-runners            Allow multiple parallel golangci-lint instances running.
                                          If false (default) - golangci-lint acquires file lock on start.
      --allow-serial-runners              Allow multiple golangci-lint instances running, but serialize them around a lock.
                                          If false (default) - golangci-lint exits with an error if it fails to acquire file lock on start.
      --path-prefix string                Path prefix to add to output
      --show-stats                        Show statistics per linter (default true)
      --output.text.path stdout           Output path can be either stdout, `stderr` or path to the file to write to.
      --output.text.print-linter-name     Print linter name in the end of issue text. (default true)
      --output.text.print-issued-lines    Print lines of code with issue. (default true)
      --output.text.colors                Use colors. (default true)
      --output.json.path stdout           Output path can be either stdout, `stderr` or path to the file to write to.
      --output.tab.path stdout            Output path can be either stdout, `stderr` or path to the file to write to.
      --output.tab.print-linter-name      Print linter name in the end of issue text. (default true)
      --output.tab.colors                 Use colors. (default true)
      --output.html.path stdout           Output path can be either stdout, `stderr` or path to the file to write to.
      --output.checkstyle.path stdout     Output path can be either stdout, `stderr` or path to the file to write to.
      --output.code-climate.path stdout   Output path can be either stdout, `stderr` or path to the file to write to.
      --output.junit-xml.path stdout      Output path can be either stdout, `stderr` or path to the file to write to.
      --output.junit-xml.extended         Support extra JUnit XML fields.
      --output.teamcity.path stdout       Output path can be either stdout, `stderr` or path to the file to write to.
      --output.sarif.path stdout          Output path can be either stdout, `stderr` or path to the file to write to.
      --max-issues-per-linter int         Maximum issues count per one linter. Set to 0 to disable (default 50)
      --max-same-issues int               Maximum count of issues with the same text. Set to 0 to disable (default 3)
      --uniq-by-line                      Make issues output unique by line (default true)
  -n, --new                               Show only new issues: if there are unstaged changes or untracked files, only those changes are analyzed, else only changes in HEAD~ are analyzed.
                                          It's a super-useful option for integration of golangci-lint into existing large codebase.
                                          It's not practical to fix all existing issues at the moment of integration: much better to not allow issues in new code.
                                          For CI setups, prefer --new-from-rev=HEAD~, as --new can skip linting the current patch if any scripts generate unstaged files before golangci-lint runs.
      --new-from-rev REV                  Show only new issues created after git revision REV
      --new-from-patch PATH               Show only new issues created in git patch with file path PATH
      --new-from-merge-base string        Show only new issues created after the best common ancestor (merge-base against HEAD)
      --whole-files                       Show issues in any part of update files (requires new-from-rev or new-from-patch)
      --fix                               Fix found issues (if it's supported by the linter)
      --cpu-profile-path string           Path to CPU profile output file
      --mem-profile-path string           Path to memory profile output file
      --print-resources-usage             Print avg and max memory usage of golangci-lint and total time
      --trace-path string                 Path to trace output file

Global Flags:
      --color string   Use color when printing; can be 'always', 'auto', or 'never' (default "auto")
  -h, --help           Help for a command
  -v, --verbose        Verbose output
```

fredrikaverpil:
@thall it's worth noting that even if you exclude paths in the config, the directories are still traversed.

From the [config file docs](https://golangci-lint.run/usage/configuration/#config-file):

```yml
linters:
  exclusions:

    # Which file paths to exclude: they will be analyzed, but issues from them won't be reported.
    # "/" will be replaced by the current OS file path separator to properly work on Windows.
    # Default: []
    paths:
      - ".*\\.my\\.go$"
      - lib/bad.go
    # Which file paths to not exclude.
    # Default: []
    paths-except:
      - ".*\\.my\\.go$"
      - lib/bad.go
```

thall:
@fredrikaverpil I understand why its failing, the problem here is that the config is located within Sage, and thus not possible to modify in order to achieve the same behavior as today.
As I see it, either all consumers needs their own config or that it's possible to pass arguments that then are merged with the default config in Sage.
Do you see other alternatives?

fredrikaverpil:
@thall no, I see it the same way as you do, unless you are fine with skipping those exclusions.

kristofferwanglund (commented):
Just a thought, since we have so much Go code and this changes both how the tool itself works and linting results, what do you think about adding a new package `sggolangcilintV2` and let everyone migrate at their own pace?

fredrikaverpil:
@kristofferwanglund yes, I thought about this too, actually. I think the bigger question is how we want to deal with the config file which might no longer be suitable for all projects if different projects needs different exclusion filters.

fredrikaverpil:
I've updated the implementation to leverage templating. Example execution:

```go
func GoLint(ctx context.Context) error {
	sg.Logger(ctx).Println("linting Go files...")
	return sggolangcilint.Run(ctx, sggolangcilint.YAMLConfigData{
		RelativePathMode:      sggolangcilint.RelativePathModeGitRoot,
		LinterExclusionsPaths: []string{".git", "node_modules$", "/tmp"},
	})
}
```

Let's see if this allows for the per-project customization we are looking for.

tools/sggolangcilint/tools.go
```diff
@@ -10,41 +10,63 @@ import (
...
 	}
 	cmdArgs := append([]string{"run", "--allow-parallel-runners", "-c", configPath}, args...)
-	cmd := Command(ctx, cmdArgs...)
+	cmd := Command(ctx, config, cmdArgs...)
 	cmd.Dir = directory
 	return cmd
 }
 
 // Run GolangCI-Lint in every Go module from the root of the current git repo.
-func Run(ctx context.Context, args ...string) error {
```
thall, line 51: This will be a breaking change for all clients, would be nice if clients that are using the default config today, doesnt need to do anything, and for clients that are using custom config, can supply a config.
Could maybe add `RunWithConfig`?

fredrikaverpil: I guess that would mean adding a new function for each existing function, right?
That would be `RunWithConfig`, `FixWithConfig`, `CommandWithConfig`, `CommandInDirectoryWithConfig`, `PrepareCommandWithConfig`.

And then I guess you would want the old `Run`, `Fix` etc hard-code a config that would be used, or did you want to keep using the old config file?

Technically, the bump to v2 _is_ a breaking change also for the CLI arguments which clients may be passing. We could add a whole new `sggolangcilintv2` package instead, and you can keep using this `sggolangcilint` (v1) as-is for as long as you want?

fredrikaverpil: I've now created a `sggolangcilintv2` package instead. Let me know what you think.

thall: I think that is better, then clients can choose when to update :+1: So marking it as deprecated is good, so we eventually can remove it.

fredrikaverpil: Ok, now I have also deprecated `sggolangcilint`. Let me know if you are missing anything.

thall: Tested it in pod-gateway, then it fails with deprecation warning :man_facepalming: maybe best to skip that initially then?
```
[sggolangcilint:prepare-command] fetching https://github.com/golangci/golangci-lint/releases/download/v1.64.8/golangci-lint-1.64.8-linux-amd64.tar.gz ...
[go-lint] main.go:132:9: SA1019: sggolangcilint.Run is deprecated: Use `sggolangcilintv2.Run` instead. Run GolangCI-Lint in every Go module from the root of the current git repo. (staticcheck)
[go-lint]       return sggolangcilint.Run(
```

fredrikaverpil: Yeah, that was expected 😄 but sure, I removed that last commit now.

radhus (approved):
I think it's time to get this in and start testing! Great work! ⭐ 

Maybe we can consider the v2 target experimental (could add a comment to be explicit) so that we're comfortable with changing the interface during our internal rollout?

PR #664, opened 2025-03-27
feat: downgrade balena

christka1:


christka1:
Found workaround, downgrade not needed

PR #668, opened 2025-04-04
feat: add uv

fredrikaverpil:
### Why?

Uv is a lot faster than poetry and comes with more functionality built-in. One particular feature is it downloads the Python version the project requires for you.

Also, since sgpoetry is installed by sgpython which reuses the local python installation, we don't get deterministic installations. This has caused issues for newly onboarded team members. Uv manages python installations, using uv finally gives us much better control in this regard.

### What?

Add uv!

### Notes

- In draft until we know we want it like this.
- Docs: https://docs.astral.sh/uv/
- Add it to your project:
	```go
	// .sage/sagefile.go
	
	func UvVersion(ctx context.Context) error {
		sg.Logger(ctx).Println("checking uv version")
		return sguv.Command(ctx, "version").Run()
	}
	```
- Works on my machine™

### DIFF
--- a/tools/sguv/command.go
+++ b/tools/sguv/command.go
@@ -1,0 +1,90 @@
+package sguv
+
+import (
+	"context"
+	"fmt"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+
+	"go.einride.tech/sage/sg"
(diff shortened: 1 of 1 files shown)


anderssonclaes (approved):
LGTM!

PR #671, opened 2025-04-25
feat(sggolangcilintv2): add formatting

fredrikaverpil:
### Why?

Golangci-lint v2 offers Go file formatting (docs/details on this [here](https://golangci-lint.run/usage/formatters/)), which is nice, as we don't have to chain the different formatters manually, or for example run only `golines` as formatter (which was the case in this repo). This can instead be delegated to golangci-lint and configured in the config yaml.

I think that this could nicely formalize how we expect formatting in CI.

### What?

- add formatters to config file template along with "sane and default" config values
- add functions for formatting; `sggolangcilintv2.Fmt` which writes the files.
- replace golines as formatter with `sggolangcilintv2.Fmt` here in the sage repo.

### Notes

- This PR supersedes https://github.com/einride/sage/pull/662
- Add argument `--diff` if you want `sggolangcilintv2.Fmt` to not write the files and return a non-zero status code when formatting is needed.

### DIFF
--- a/.sage/main.go
+++ b/.sage/main.go
@@ -14,7 +14,6 @@ import (
 	"go.einride.tech/sage/tools/sggo"
 	"go.einride.tech/sage/tools/sggolangcilintv2"
 	"go.einride.tech/sage/tools/sggolicenses"
-	"go.einride.tech/sage/tools/sggolines"
 	"go.einride.tech/sage/tools/sggopls"
 	"go.einride.tech/sage/tools/sgmdformat"
 	"go.einride.tech/sage/tools/sgyamlfmt"
@@ -73,7 +72,10 @@ func GoLintFix(ctx context.Context) error {
 
 func GoFormat(ctx context.Context) error {
 	sg.Logger(ctx).Println("formatting Go files...")
-	return sggolines.Run(ctx)
+	return sggolangcilintv2.Fmt(
+		ctx,
+		sggolangcilintv2.Config{RunRelativePathMode: sggolangcilintv2.RunRelativePathModeGitRoot},
+	)
 }
 
 func GoLicenses(ctx context.Context) error {
--- a/tools/sggolangcilintv2/tools.go
+++ b/tools/sggolangcilintv2/tools.go
@@ -43,6 +43,12 @@ type Config struct {
 	// - allow you to optionally use regexps here, like ".*\\.my\\.go$".
 	// - treat paths relative to the setting of RunRelativePathMode.
 	LintersExclusionsPaths []string
+	// Which file paths to exclude from issue reporting. The paths will still be analyzed.
+	// Golangci-Lint will:
+	// - replace "/" with the current OS file path separator to properly work on Windows.
+	// - allow you to optionally use regexps here, like ".*\\.my\\.go$".
+	// - treat paths relative to the setting of RunRelativePathMode.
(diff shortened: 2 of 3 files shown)


tools/sggolangcilintv2/tools.go
```diff
@@ -119,6 +125,57 @@ func Fix(ctx context.Context, config Config, args ...string) error {
...
+// Run `golangci-lint fmt --diff` in every Go module from the root of the current git repo.
+func Fmt(ctx context.Context, config Config, args ...string) error {
+	var commands []*exec.Cmd
+	if err := filepath.WalkDir(sg.FromGitRoot(), func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if d.IsDir() || d.Name() != "go.mod" {
+			return nil
+		}
```
fredrikaverpil, line 137: Hey @joaoeinride you [commented](https://github.com/einride/sage/pull/662/files#r2014017145) in an old draft PR about this. Figured so this doesn't get lost in this new PR that I'd just ping you about it.

I did not quite understand why you'd want to change the condition or check for `.go` files.

The idea is to run `golangci-lint run -c <config-file>` in every location where there is a `go.mod` file, recursively from the git root.

Please let me know if you think we should change this approach. I've really just copied the whole setup from `sggolanglint` (v1) and there this `WalkDir` + if-condition has been used for all the different functions.

tools/sggolangcilintv2/tools.go
```diff
@@ -119,6 +125,57 @@ func Fix(ctx context.Context, config Config, args ...string) error {
 	return nil
 }
 
+// Run `golangci-lint fmt --diff` in every Go module from the root of the current git repo.
+func Fmt(ctx context.Context, config Config, args ...string) error {
```
radhus, line 129: We previously just provided a target that always wrote files like `FmtFix`. As we typically run git to verify no diff in the end, I think it's find to let `Fmt` write (i.e. be `FmtFix`). No strong opinion though, but less code to maintain 🤓

.sage/main.go
```diff
@@ -73,7 +72,10 @@ func GoLintFix(ctx context.Context) error {
 
 func GoFormat(ctx context.Context) error {
 	sg.Logger(ctx).Println("formatting Go files...")
-	return sggolines.Run(ctx)
+	return sggolangcilintv2.Fmt(
```
radhus, line 75: Corresponding function should be `FixFix` as `sggolines.Run` wrote changes.
```suggestion
	return sggolangcilintv2.FmtFix(
```

radhus (approved):
Nice

tools/sggolangcilintv2/tools.go
joaoeinride: gotcha, that makes sense 👍🏻

.sage/main.go
fredrikaverpil: Done!

tools/sggolangcilintv2/tools.go
fredrikaverpil: Right, makes sense. You can also just pass the `--diff` argument if you don't want it to write any files. 👍

fredrikaverpil: Fixed!

PR #675, opened 2025-05-12
feat(sggolangcilint): add support for empty go.mod files

joaoeinride:
according to docs https://go.dev/ref/mod#vcs-dir

```For example, if a repository has large files checked into a testdata directory,
the module author could add an empty go.mod file in testdata so their users don’t
need to download those files.
```

This adds support for sggolangcilint module to ignore folders containing empty go.mod files.


thall (approved):

radhus (approved):

joaoeinride:
will do, but shouldn't `sggolangcilintv2` be in a v2 folder?

PR #678, opened 2025-05-16
perf(ci/setup-sage): include golangci-lint cache in sage cache

kristofferwanglund:
In one of our repos this made the time linting go down from about `1min` to `4s`.


fredrikaverpil (approved):
Nice!

PR #690, opened 2025-09-16
feat(sggolangcilintv2): add ability to run formatter per-directory

fredrikaverpil:


alethenorio:
Had a chat with @fredrikaverpil and we felt that some renaming might be in place for this package as it fairly new to make the naming a bit more clear

PR #699, opened 2025-09-30
fix(action): inputs only takes strings

fredrikaverpil:
### Why?

- Apparently we are setting non-string types for inputs.

### What?

- Change input defaults into string types. They are already treated as strings
  in the if-conditions etc.

### DIFF
--- a/actions/setup/action.yml
+++ b/actions/setup/action.yml
@@ -10,7 +10,7 @@ inputs:
   disableCache:
     description: Disable cache
     required: false
-    default: false
+    default: 'false'
 
   go-version:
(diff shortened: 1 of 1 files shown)


alethenorio (commented):
Haven't these been liked this for a while? If that's the case it means GitHub clearly accepts them? So maybe we should rely on GitHub doing the conversion instead of us doing so?

fredrikaverpil:
> Haven't these been liked this for a while? If that's the case it means GitHub clearly accepts them? So maybe we should rely on GitHub doing the conversion instead of us doing so?

Yes. I don't think it matters other than that this PR "makes it right" and avoids potential confusion and linter warnings, like this one:

image

PR #712, opened 2025-10-16
ci: remove gopls from ci

fredrikaverpil:
We don't really use this CI workflow for anything. Recently, it started
failing and was fixed in https://github.com/einride/sage/pull/711

To avoid unnecessary noise and maintenance, I'm removing it from CI
here.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -18,18 +18,3 @@ jobs:
 
       - name: Make
         run: make
-
-  gopls:
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v4
-
-      - name: Setup Sage
-        uses: ./actions/setup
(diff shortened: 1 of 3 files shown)


.github/workflows/go.yml
```diff
@@ -18,18 +18,3 @@ jobs:
...
-      - name: Checkout
-        uses: actions/checkout@v4
-
-      - name: Setup Sage
-        uses: ./actions/setup
-        with:
-          go-version: "stable"
-
-      - name: Make
-        run: make go-pls
```
fredrikaverpil: @joaoeinride hey I know you were interested in running this in your own CI, but we are removing it from here to avoid unnecessary churn. Hope you're ok with that 😄

joaoeinride: that's the same reason we haven't added it on our side 😆

PR #715, opened 2025-10-30
feat(sgapilinter): bump to v2.0.0

fredrikaverpil:


alethenorio (changes requested):
This needs further refining. See https://github.com/einride/sage/pull/716

PR #723, opened 2025-11-05
feat: bump required go version to 1.24.9

fredrikaverpil:
### Why?

- We should bump Go as we follow the same release policy as [Go's own release policy](https://go.dev/doc/devel/release#policy).
- We cannot upgrade certain dependencies, such as GoLicenses to its latest version.

```
[sggolicenses:prepare-command] go: github.com/google/go-licenses/v2@v2.0.1: github.com/google/go-licenses/v2@v2.0.1 requires go >= 1.23.0 (running go 1.22.12; GOTOOLCHAIN=local)
```

### What?

- Bump required Go version for Sage. Rather than bumping to 1.23, we are bumping to 1.24.
- Remove GOROOT (`GoLicenses` has been validated to still work as expected in one internal project):
  ```
	Error: tools/sggolicenses/tools.go:56:53: SA1019: runtime.GOROOT has been deprecated since Go 1.24: The root used during the Go build will not be meaningful if the binary is copied to another machine. Use the system path to locate the “go” binary, and use “go env GOROOT” to find its GOROOT. (staticcheck)
	cmd.Env = append(cmd.Env, fmt.Sprintf("GOROOT=%s", runtime.GOROOT()))
  ```

### DIFF
--- a/.sage/go.mod
+++ b/.sage/go.mod
@@ -1,6 +1,6 @@
 module github.com/einride/sage/.sage
 
-go 1.22
+go 1.24.9
 
 require go.einride.tech/sage v0.0.0-00010101000000-000000000000
 
--- a/go.mod
+++ b/go.mod
@@ -1,3 +1,3 @@
 module go.einride.tech/sage
 
-go 1.22
+go 1.24.9
--- a/tools/sggolicenses/tools.go
+++ b/tools/sggolicenses/tools.go
@@ -8,7 +8,6 @@ import (
 	"io/fs"
 	"os/exec"
 	"path/filepath"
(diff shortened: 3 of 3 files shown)


ericwenn:
> @ericwenn You're correct. However, in this case, we actually also want to follow the release policy (we follow [Go's release policy](https://go.dev/doc/devel/release#policy)). I forgot to mention this in the PR description... (added it now). Do you still feel that we should not move ahead with the proposed change of bumping `go` (not `toolchain`)?

@fredrikaverpil Thanks for links to release policy, I was not aware of that. Lets go ahead with this change!

PR #734, opened 2025-11-27
feat(sgbuf): bump to v1.61.0

fredrikaverpil:
Bumps buf from 1.59.0 to 1.61.0.

Repository: https://github.com/bufbuild/buf


fredrikaverpil:
note that this contains all the commits from #735 too... which will not be the case later.

PR #816, opened 2026-05-12
feat(sggolines): move to golangci/golines

loodde:
segmentio/golines is archived and directs to golangci/golines instead which is a fork of segmentio, this commit migrates to use the golangci fork instead.

### DIFF
--- a/tools/sggolines/tools.go
+++ b/tools/sggolines/tools.go
@@ -14,8 +14,9 @@ import (
 )
 
 const (
-	name    = "golines"
-	version = "0.12.2"
+	name = "golines"
+	// renovate: datasource=github-releases depName=golangci/golines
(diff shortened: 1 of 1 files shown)


fredrikaverpil:
Hey @loodde, I see you closed this. Should we use golines via golangci-lint instead and deprecate this one?

loodde:
@fredrikaverpil I saw we already have golines as a formatter in golangcilint yaml conf. So think that should cover the need. Regarding deprecation of this I guess that is more of a question of how we handle deprecations in sage? I could not find any users of this tool directly in the einride org other than sage itself.

fredrikaverpil:
@loodde ok sounds good. 👍 I think we can deprecate this one. I can open a PR.

PR #817, opened 2026-05-12
chore(sggolines): deprecate package

fredrikaverpil:
## Why?

golangci-lint v2 ships with a built-in golines formatter, 
a fork of the original golines. So maintaining a
standalone sggolines tool is no longer necessary.

## What?

- Add package-level `Deprecated:` notice on `tools/sggolines`
- Add function-level `Deprecated:` notices on `Run`, `Command`, and
  `PrepareCommand`, pointing users at the corresponding `sggolangcilint`
  equivalents

### DIFF
--- a/tools/sggolines/tools.go
+++ b/tools/sggolines/tools.go
@@ -1,3 +1,7 @@
+// Deprecated: sggolines is deprecated. golangci-lint v2 includes a built-in
+// golines formatter, so a standalone tool is no longer needed.
+//
+// See sggolangcilint for a replacement.
 package sggolines
 
 import (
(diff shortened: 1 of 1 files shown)


tools/sggolines/tools.go
```diff
@@ -19,6 +23,8 @@ const (
 )
 
 // Run golines on all Go files in the current git root with gofumpt as default formatter.
+//
+// Deprecated: Use sggolangcilint.Run instead; golangci-lint v2 has a built-in golines formatter.
```
loodde, line 27: I guess technically is is not built in but rather it must be specified as an enabled formatter in the golangcilint yaml file, looks like this currently:

```
formatters:
  enable:
    - gci
    - goimports
    - gofumpt
    - golines
```

Edholm (approved):

tools/sggolines/tools.go
fredrikaverpil: Yeah, golines is enabled by default in the config file we bundle with golangci-lint v2.

PR #818, opened 2026-05-13
fix(sg/initfile): swap places of arguments 'got' and 'expected'

JoakimKjellberg:
The printout of `wrong number of arguments to %s, got %v expected %v` had mixed up the values for "got" and "expected".


Edholm (approved):
nice catch!

PR #823, opened 2026-06-03
feat(sgpostgres): allow configuring postgres version

fredrikaverpil:
## Why?

- `RunLocal` hard-coded the postgres image version (`14`), so callers could not run tests against other versions.

## What?

- Add a `WithVersion` functional option to `RunLocal`.
- Default to `14` when no option is supplied.
- Resolve the image tag at call time instead of via a package-level constant.


kristofferwanglund (approved):
LGTM

PR #853, opened 2026-08-11
fix: use `--ignore-scripts` flag for npm install

Edholm:
npm 12 no longer accepts the singular `--ignore-script` alias.


fredrikaverpil (approved):
👍

PR #854, opened 2026-08-12
fix(sgcloudrun): support service account source credentials in LocalDevelop

alethenorio:
LocalDevelopCommand could only impersonate a service account from a local
user credential. Three issues prevented it from working when the caller was
itself authenticated as a service account, for example using the
google-github-actions/auth:

- Application Default Credentials were read only from gcloud's well-known
  path. Instead, resolve GOOGLE_APPLICATION_CREDENTIALS and
  CLOUDSDK_AUTH_CREDENTIAL_FILE_OVERRIDE first, matching the Google Cloud
  client libraries, and support CLOUDSDK_CONFIG for the well-known path.
- Self-delegation through the impersonated service account was applied
  unconditionally. Restrict it to authorized_user source credentials; service
  account and external account credentials use plain impersonation, which only
  needs roles/iam.serviceAccountTokenCreator on the target rather than also
  requiring the target to hold that role on itself.
- The generated impersonated credentials file was silently overridden. The
  process environment was appended after it and os/exec keeps the last
  occurrence of a duplicated key, so an ambient GOOGLE_APPLICATION_CREDENTIALS
  won and the service ran as the caller's identity with no error. Exclude the
  credential variables from the override pass when impersonating.

Also mint the source access token from the resolved credentials file so the
token identity cannot diverge from the wrapped source credentials, and
document the resolution order and IAM requirement on LocalDevelopCommand.

Note that resolving the environment variables before gcloud's well-known path
changes which credentials are used for developers who export
GOOGLE_APPLICATION_CREDENTIALS locally.

### DIFF
--- a/tools/sgcloudrun/config.go
+++ b/tools/sgcloudrun/config.go
@@ -5,7 +5,6 @@ import (
 	"context"
 	"encoding/base64"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io"
 	"math/rand/v2"
@@ -13,6 +12,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"slices"
 	"strings"
 
 	"go.einride.tech/sage/sg"
@@ -92,37 +92,13 @@ func LocalDevelopEnv(
 	projectID string,
 	serviceAccountEmail string,
 ) ([]string, error) {
-	// Grab the local user token to impersonate the service account
-	currentADC, err := applicationDefaultCredentials()
+	creds, err := resolveCredentials()
 	if err != nil {
 		return nil, err
 	}
+	delegates := impersonationDelegates(creds.credentialType, serviceAccountEmail)
 
-	// Store a local token wrapping the user token in metadata to make a delegated request.
-	// This requires the impersonated service account to have implicitDelegation permissions on itself.
-	// See https://cloud.google.com/iam/docs/create-short-lived-credentials-delegated#sa-credentials-delegated for details
-	serviceAccountImpersonationURL := fmt.Sprintf(
-		"https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/%s:generateAccessToken",
-		serviceAccountEmail,
-	)
-
-	delegateCreds := struct {
-		Delegates                      []string        `json:"delegates"`
-		Type                           string          `json:"type"`
-		ServiceAccountImpersonationURL string          `json:"service_account_impersonation_url"`
-		SourceCredentials              json.RawMessage `json:"source_credentials"`
-	}{
-		Delegates:                      []string{"projects/-/serviceAccounts/" + serviceAccountEmail},
-		Type:                           "impersonated_service_account",
-		ServiceAccountImpersonationURL: serviceAccountImpersonationURL,
-		SourceCredentials:              json.RawMessage(currentADC),
-	}
-	delegateCredsJSON, err := json.Marshal(delegateCreds)
-	if err != nil {
-		return nil, err
-	}
-
-	accessToken, err := fetchImpersonatedAccessToken(ctx, serviceAccountEmail)
+	accessToken, err := fetchImpersonatedAccessToken(ctx, creds, serviceAccountEmail, delegates)
 	if err != nil {
 		return nil, fmt.Errorf("unable to fetch impersonated service account access token: %v", err)
 	}
@@ -132,19 +108,16 @@ func LocalDevelopEnv(
 		return nil, err
 	}
 
-	workDir := sg.FromBuildDir("gcloud")
-	if err := os.MkdirAll(workDir, 0o755); err != nil {
-		return nil, fmt.Errorf("unable to create path to store gcloud credentials: %v", err)
-	}
-	credsPath := filepath.Join(workDir, fmt.Sprintf("creds-%s.json", randomLower(5)))
-	if err := os.WriteFile(credsPath, delegateCredsJSON, 0o600); err != nil {
+	// The temporary credentials file is removed by CleanUpLocalDevelop, so the cleanup func is not needed here.
+	credsPath, _, err := delegatedGoogleApplicationCredentials(creds, serviceAccountEmail, delegates)
+	if err != nil {
 		return nil, err
 	}
 
 	env = append(env, "K_REVISION=local"+sggit.SHA(ctx))
 	env = append(env, "K_CONFIGURATION="+configFile)
 	env = append(env, "GOOGLE_CLOUD_PROJECT="+projectID)
-	env = append(env, "GOOGLE_APPLICATION_CREDENTIALS="+credsPath)
+	env = append(env, googleApplicationCredentialsEnvVar+"="+credsPath)
 
 	return env, nil
 }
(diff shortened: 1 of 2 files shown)


tools/sgcloudrun/config.go
```diff
@@ -383,28 +367,157 @@ func accessSecretVersion(ctx context.Context, accessToken, project, secret, vers
...
+		configDir = filepath.Join(home, ".config", "gcloud")
+	}
+	candidates = append(candidates, candidate{path: filepath.Join(configDir, "application_default_credentials.json")})
+	tried := make([]string, 0, len(candidates))
+	for _, c := range candidates {
+		tried = append(tried, c.path)
+		data, err := os.ReadFile(c.path)
+		if err != nil {
+			if os.IsNotExist(err) {
+				continue
```
mvestin, line 448: If the file is missing, do we really always want to fall back to the other candidates? If it is e.g. set through a miss-spelled env var, we can end up with the same type of unexpected impersonation as the ones we are fixing in this PR.

Maybe we should return an error instead, if c.fromEnv is true?

tools/sgcloudrun/config.go
```diff
@@ -383,28 +367,157 @@ func accessSecretVersion(ctx context.Context, accessToken, project, secret, vers
...
+		Type string `json:"type"`
 	}
-	path := filepath.Join(home, ".config/gcloud/application_default_credentials.json")
-	b, err := os.ReadFile(path)
-	if err != nil {
-		if os.IsNotExist(err) {
-			return "", errors.New(
-				"no application default credentials found. Please authenticate using 'gcloud auth application-default login'",
+	if err := json.Unmarshal(c.data, &document); err != nil {
+		return ""
```
mvestin: Would it be safer to return an "Invalid JSON error" instead of empty string?

tools/sgcloudrun/config.go
```diff
@@ -155,21 +128,29 @@ func LocalDevelopEnv(
 // Deprecated: There was no reason to export this function and it may get removed in the future.
 func CleanUpLocalDevelop(environ []string) error {
 	for _, env := range environ {
-		if !strings.Contains(env, "GOOGLE_APPLICATION_CREDENTIALS") {
+		if !strings.Contains(env, googleApplicationCredentialsEnvVar) {
```
mvestin, line 131: Should this be `googleApplicationCredentialsEnvVar+"="`  (similar to `withoutEnvVars`)?

alethenorio: Yeah we could as well but not sure it would make a difference (plus I am avoiding adding any potential side effects to a deprecated public function)

tools/sgcloudrun/config.go
alethenorio: I've refactored this a bit and the fromEnv variable is gone and replaced with just checking envVar and we return an error if the envVar is set and we're unable to read file

tools/sgcloudrun/config.go
alethenorio: I've refactored this and now the unmarshal returns an error as well

mvestin (approved):

tools/sgcloudrun/config.go
fredrikaverpil, line 86: Out of scope for this change, but it seems like it would've been a good idea to print this in the terminal. I don't think most users actually read this source code before executing each `make` command.

This note occurs in a few places in this file and applies to several targets.

tools/sgcloudrun/config.go
```diff
@@ -383,28 +367,174 @@ func accessSecretVersion(ctx context.Context, accessToken, project, secret, vers
...
+// A path named by an environment variable must exist and hold valid JSON. Falling back to the well-known path
+// when it does not would silently impersonate from whatever credentials happen to be there, which is the class
+// of surprise this resolution order exists to avoid.
+func resolveCredentials() (resolvedCredentials, error) {
+	type candidate struct {
+		path string
+		// envVar is the environment variable the path came from, empty for gcloud's well-known path.
+		envVar string
+	}
+	candidates := make([]candidate, 0, 3)
```
fredrikaverpil: Why `3`?
Reads as if you had something specific in mind there.

tools/sgcloudrun/config.go
```diff
@@ -383,28 +367,174 @@ func accessSecretVersion(ctx context.Context, accessToken, project, secret, vers
...
+// reach it without a direct getAccessToken binding. See
+// https://cloud.google.com/iam/docs/create-short-lived-credentials-delegated#sa-credentials-delegated
+// Service account and external account source credentials use plain impersonation instead: they only need
+// roles/iam.serviceAccountTokenCreator on the target, whereas self-delegation would additionally require the
+// target service account to hold that role on itself.
+func impersonationDelegates(credentialType, serviceAccountEmail string) []string {
+	if credentialType != credentialTypeAuthorizedUser {
+		return nil
+	}
+	return []string{"projects/-/serviceAccounts/" + serviceAccountEmail}
```
fredrikaverpil, line 493: So if my understanding is correct... we could add several delegates here (which would cause the POST request to fail if each delegate could not impersonate the other one in the chain) when minting a new impersonated token, correct?

It could then be kind of similar as to doing something like the below, right?

```sh
gcloud auth print-access-token \
	--impersonate-service-account=SA_TARGET,SA_DELEGATE_1,SA_DELEGATE_2
```

alethenorio: Kind of. The last element is supposedly the target in the gcloud example and the ones before are the delegates but otherwise I think your logic is correct

fredrikaverpil: Ah, maybe that was a little unclear. This is not something we need to get stuck on, but what I meant was:

- SA_TARGET: the impersonated SA. The command prints an access token that is this account.
- SA_DELEGATE_1, SA_DELEGATE_2: the intermediaries the authorization hops through; they only affect which IAM bindings gets checked/validated so to reach to the SA_TARGET.

Meaning; logically `you -> SA_DELEGATE_1 -> SA_DELEGATE_2 -> SA_TARGET`.

From the docs:

> For example, when --impersonate-service-account= SERVICE_ACCOUNT_1,SERVICE_ACCOUNT_2, the active account must have the roles/iam.serviceAccountTokenCreator role on SERVICE_ACCOUNT_1, which must have the roles/iam.serviceAccountTokenCreator role on SERVICE_ACCOUNT_2. SERVICE_ACCOUNT_1 is the impersonated service account and SERVICE_ACCOUNT_2 is the delegate. 

https://docs.cloud.google.com/sdk/gcloud/reference#--impersonate-service-account

fredrikaverpil (approved):
Lots of moving parts, quite complex – but especially since it's been evaluated and works, LGTM!

tools/sgcloudrun/config.go
alethenorio: I'm not sure a log statement is suited here but let's get back to it to not block this PR

fredrikaverpil: Yes, it was mostly something I noticed. I'm also not sure whether these deprecated commands are actually used anywhere. Might be better to e.g. make them internal or remove them in a while.

PR #867, opened 2026-08-31
fix(sggolangcilintv2): skip dir instead of skip all on empty go.mod file

JoakimKjellberg:
When `SkipAll` is returned, all the directories after this one are skipped, meaning that the linting is aborted. Using `SkipDir` instead only skips the current directory and its sub-directories, but not the ones after it.

[Slack thread](https://einride.slack.com/archives/CA484JN22/p1788181356362579) with some background.


Edholm (approved):
Huh nice find!

PR #868, opened 2026-08-31
fix(sggolangcilintv2): don't prune sibling entries on empty go.mod

fredrikaverpil:
## Why?

- Follow-up to #867: [`filepath.SkipDir`](https://pkg.go.dev/path/filepath#SkipDir) returned while visiting a *file* does not skip that directory's subtree — per the [`io/fs.WalkDirFunc` docs](https://pkg.go.dev/io/fs#WalkDirFunc) it skips the remaining entries of the *parent* directory, and [`filepath.WalkDir`](https://pkg.go.dev/path/filepath#WalkDir) walks in lexical order.
- Result: which nested modules get linted depends on whether their directory name sorts before or after the string `go.mod`:

```
root/
├── api/
│   └── go.mod        <- linted ("api" < "go.mod", walked before SkipDir)
├── go.mod            <- empty, returns SkipDir
└── pkg/
    └── go.mod        <- silently skipped ("pkg" > "go.mod", never visited)
```

## What?

- Return `nil` instead of `filepath.SkipDir` on an empty `go.mod`, ignoring only that file — every other directory is still walked and its module linted, the same handling `Fix()` and `Fmt()` already use.

## Notes

- Verified with a minimal `WalkDir` reproduction of the layout above: `api` was linted, `pkg` was not; renaming `api` to `zapi` made zero modules lint.

### DIFF
--- a/tools/sggolangcilintv2/tools.go
+++ b/tools/sggolangcilintv2/tools.go
@@ -100,9 +100,9 @@ func Run(ctx context.Context, config Config, args ...string) error {
 		if err != nil {
 			return err
 		}
-		// ignore this directory and its sub-directories if it's an empty go.mod
+		// ignore if it's an empty go.mod
 		if fileInfo.Size() == 0 {
-			return filepath.SkipDir
+			return nil
 		}
 		cmd := CommandRunInDirectory(ctx, config, filepath.Dir(path), args...)
 		commands = append(commands, cmd)


tools/sggolangcilintv2/tools.go
```diff
@@ -100,9 +100,9 @@ func Run(ctx context.Context, config Config, args ...string) error {
 		if err != nil {
 			return err
 		}
-		// ignore this directory and its sub-directories if it's an empty go.mod
+		// ignore if it's an empty go.mod
 		if fileInfo.Size() == 0 {
-			return filepath.SkipDir
+			return nil
```
alethenorio, line 105: If I am not mistaken, an empty go.mod is used for very specific use cases but never as an actual go module so I think ignoring the directory here might be correct as we should not expect actual go.mod files inside an empty go.mod directory.

alethenorio: For completion, this is the case for an empty go.mod that I can see as documented here

https://go.dev/ref/mod#vcs-dir

alethenorio: It also looks like empty go.mod is completely [ignored](https://go.dev/wiki/Modules#can-an-additional-gomod-exclude-unnecessary-content-do-modules-have-the-equivalent-of-a-gitignore-file) by Go so I am not sure we should be linting anything inside it

fredrikaverpil: Right, hmm. That's an argument against both the merged code and this PR, because SkipDir-at-the-file doesn't implement that behavior either.

alethenorio: We still should not fail on an empty go.mod though (which I believe is what the previous PR was trying to fix?)

fredrikaverpil: The merged code (now in master) introduced:

```go
if fileInfo.Size() == 0 {
	return filepath.SkipDir
}
```

Meaning, it would skip the current dir being parsed, which could lead to this problem:

```
root/
├── api/
│   └── go.mod        <- linted ("api" < "go.mod", walked before SkipDir)
├── go.mod            <- empty, returns SkipDir
└── pkg/
    └── go.mod        <- silently skipped ("pkg" > "go.mod", never visited)
```

alethenorio: But that's OK. As the links above show, Go will itself ignore an empty go.mod and all data under it so if there is actual data to be limited in a folder under an empty go.mod then maybe that's a mistake?

JoakimKjellberg: @alethenorio, regarding the previous PR: the linting did not fail on an empty go.mod - it completely stopped linting when it reached one. Meaning that it ended with success even though the directories that I wanted to lint actually contained errors.

PR #871, opened 2026-09-10
feat(renovate): add dependencies label and force feat commit type for all updates

alethenorio:
feat(renovate): add dependencies label and force feat commit type for all updates

Adding a "dependencies" label to all renovate PRs. This brings it in line with how dependabot works out of the box and allows for similar functionality such as filtering on all PRs with that label.

Also, "chore" does not trigger a release by default in most semantic release tooling and dependency changes are potential code changes so we switch from "chore" to "feat" instead.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -4,7 +4,7 @@
   "minimumReleaseAge": "3 days",
   "postUpdateOptions": ["gomodTidy"],
   "semanticCommits": "enabled",
-  "semanticCommitType": "feat",
+  "labels": ["dependencies"],
   "packageRules": [
     {
       "matchManagers": [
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -5,6 +5,7 @@
   "postUpdateOptions": ["gomodTidy"],
   "semanticCommits": "enabled",
   "semanticCommitType": "feat",
```
fredrikaverpil: I knew I tried to fix this at some point!
This is likely moot then, and can be removed?

```suggestion
```

fredrikaverpil (approved):

renovate.json
alethenorio: Removed it.

PR #875, opened 2026-09-14
Drop markdown formatting from default target

Tethik:
## Summary
Removes `FormatMarkdown` from the default target in both sage's own `.sage/main.go` and the `example/.sage/main.go` template embedded into `sage init`. mdformat is brittle and its reformatting churn costs unnecessary compute.

## Notes
The `sgmdformat` tool package is untouched — repos that want it can still wire it up explicitly. Existing repos keep their current setup; this only changes what new `sage init` scaffolding generates.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Tethik:
Non-claude, 👷🏼 opinion: never found this formatter useful, and too many times it has just been annoying to set up. I do think it just also wastes compute/tokens. Could say the same about the yaml formatter tbh.

fredrikaverpil:
@Tethik we're evaluating a different formatter in one of our projects; https://rumdl.dev
So far looking promising. I'd like to eval it in more projects. Maybe yours would benefit from it as well? I'll reach out in DM.

Tethik:
Summarizing a bit what I wrote in DMs:
My issue here is that I think running a markdown formatter in the first place feels unnecessary, so I propose to make it optional by default.

Not against switching to a different formatter though, that feels like an improvement too, but I would still propose to make markdown (and perhaps also yaml) an optional plugin, not default.

fredrikaverpil:
> Summarizing a bit what I wrote in DMs:
> 
> My issue here is that I think running a markdown formatter in the first place feels unnecessary, so I propose to make it optional by default.
> 
> 
> 
> Not against switching to a different formatter though, that feels like an improvement too, but I would still propose to make markdown (and perhaps also yaml) an optional plugin, not default.

Just so it is clear and unless I'm missing something; you just removed markdown formatting here in this repo. You are free to remove it in any of your teams' repos too. It's already optional, so to speak, and up to your team on how to handle in your own projects.

Besides that note, we can of course discuss this further internally.

Tethik:
> > Summarizing a bit what I wrote in DMs:
> > My issue here is that I think running a markdown formatter in the first place feels unnecessary, so I propose to make it optional by default.
> > Not against switching to a different formatter though, that feels like an improvement too, but I would still propose to make markdown (and perhaps also yaml) an optional plugin, not default.
> 
> Just so it is clear and unless I'm missing something; you just removed markdown formatting here in this repo. You are free to remove it in any of your teams' repos too. It's already optional, so to speak, and up to your team on how to handle in your own projects.
> 
> Besides that note, we can of course discuss this further internally.

Yes, it is optional, but I think it ought to be opt-in not opt-out.*

Ah, but I see what you mean, it shouldn't change this repo's own .sage

Tethik:
Right, sorry. I get what you mean - I think I need to find the internal template instead, this is the wrong repo.

